PR #85, opened 2020-12-26
Allow also odd-order butterworth filters

megajocke:
Hi Henrik

Found your project through faktiskt.io and tried it out a bit.
Very nice! Managed to build it on Linux according to the instructions and run it with Alsa and a 5.1 channel SoundBlaster SB1095 USB sound interface.

Found a little bug though when I tried to create a fifth order butterworth filter:
The check for Butterworth filters order only allowed even-order filters.
Jag gissar att det är den ökända "cut-'n-paste-katten" som varit framme om man jämför med LR-filtrets kontroll. ;)

### DIFF
--- a/.github/workflows/ci_test.yml
+++ b/.github/workflows/ci_test.yml
@@ -7,6 +7,11 @@ jobs:
     name: Check test and lint Linux
     runs-on: ubuntu-latest
     container: ubuntu:20.04
+
+    env:
+      # for apt-get: otherwise installing tzdata gets stuck asking about the time zone.
+      DEBIAN_FRONTEND: noninteractive
+
     steps:
       - name: Checkout sources
         uses: actions/checkout@v2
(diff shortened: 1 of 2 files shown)


megajocke:
Ah, maybe I should have done the changes against the develop branch instead.

HEnquist:
Oops! Thanks for this :) This has been there since the very first release, funny that nobody noticed until now.
Seems like I had a bad day I wrote this part, the check for LR is also broken: 
(*order % 2 > 0) && (*order == 0)
Pretty useless, it can never become true!
I'll fix that too and make a 0.4.1 release.

megajocke:
I think I have done the same logic mistake and not just once...

So for LR it should be
(*order % 2 > 0) || (*order == 0)
then?

I guess it won't matter in both places if it's checked as order == 0 or <= 0 as order is unsigned if I'm not mistaken?

HEnquist:
Master is updated, and a new release 0.4.1 is up! This will also be fixed in the next beta of 0.5.0.

PR #125, opened 2021-07-30
alsa: Support Capture Pitch 1000000 rate shift

t123yh:
In Linux 5.14, a new feedback mechanism for USB Audio Gadget is introduced (commit e89bb428), along with a new control element "Capture Pitch 1000000". This patch adds support for this feature.

### DIFF
--- a/src/alsadevice.rs
+++ b/src/alsadevice.rs
@@ -374,15 +374,22 @@ fn capture_loop_bytes(
     let card = pcminfo.get_card();
     let device = pcminfo.get_device();
     let subdevice = pcminfo.get_subdevice();
-    let mut elid = ElemId::new(ElemIface::PCM);
-    elid.set_device(device);
-    elid.set_subdevice(subdevice);
-    elid.set_name(&CString::new("PCM Rate Shift 100000").unwrap());
     let h = HCtl::new(&format!("hw:{}", card), false).unwrap();
     h.load().unwrap();
-    let element = h.find_elem(&elid);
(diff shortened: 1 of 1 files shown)


HEnquist:
Thanks for this! Just one minor thing to make it slightly easier to read. Could you rename `elem` and `elem2` to `elem_loopback` and `elem_gadget`?

t123yh:
@HEnquist Updated. BTW this is not very well tested (I only confirmed that the pitch is stable), so could you please test if it works? Thanks.

HEnquist:
Thanks! It looks like it should work, but I'll try it. Do you happen to know a quick way to get kernel 5.14 running on a RPi?

t123yh:
> Thanks! It looks like it should work, but I'll try it. Do you happen to know a quick way to get kernel 5.14 running on a RPi?

Sorry but I'm testing on a Rock Pi S :) In fact, you don't have to have a 5.14 kernel to test this. You just copy u_audio.c, f_uac1.c, f_uac2.c and the respective headers from 5.14-rc2 to 5.10 kernel and it will work.

HEnquist:
Ok, I had a tiny hope that maybe I wouldn't need to compile the kernel myself :)
I'll merge this for now and test it once as soon as I have some time.

HEnquist:
I have compiled kernel 5.14 and made a quick test. So far it seems to work just fine. But I'll leave it running for a few hours to make sure.

HEnquist:
Now it's been running for much longer, and it's keeping the buffer level perfectly!

PR #162, opened 2021-11-23
Remove outdated REW script as REW no longer supports output XML.

Wang-Yue:


HEnquist:
Hmm what do you mean? I just tried and could export XML from REW 5.20.4.

Wang-Yue:
I only see the option to export "in REW .req binary format". In Export I only see export filters as wav or text. No XML. 
What version are you using? I'm using the Mac version. Maybe I found the wrong place?

HEnquist:
REW 5.20.4 on Linux. 
In the EQ window, select rePhase as equalizer, and then "Save filter coefficients to file" under "Filter Tasks" will save as XML.

Wang-Yue:
LOL... I was looking for XML in the menu and could not found it. Now I see how it works! Many thanks! Lemme close this pull request.

PR #176, opened 2022-01-17
log config file reload as info

herrernst:
I'd still like to see config reloads without running with `-v` (which floats my Terminal with sample rate watcher messages).


HEnquist:
Thanks for this, yes there should definitely be some message in info level when reloading the config. I'll just look it over and make sure it's consistent between the different ways of triggering the reload.

PR #178, opened 2022-01-19
Next100

HEnquist:


src/alsadevice.rs
```diff
@@ -161,42 +169,58 @@ fn play_buffer(
...
                 "Playback device failed while waiting for available buffer space, error: {}",
                 err
             );
-            return Err(Box::new(err));
+                return Err(Box::new(err));
+            }
         }
+
+        match io.writei(buffer) {
+            Ok(frames_written) => {
```
HEnquist, line 191: Could `frames_written` sometimes be zero? If yes, do we need to handle that so the loop doesn't spin forever?

src/alsadevice.rs
```diff
@@ -220,42 +244,53 @@ fn capture_buffer(
...
-            alsa::nix::errno::Errno::EPIPE => {
-                warn!("Capture failed, error: {}", err);
                 return Err(Box::new(err));
             }
-            _ => {
-                warn!("Capture failed, error: {}", err);
-                return Err(Box::new(err));
+        }
+        match io.readi(buffer) {
+            Ok(frames_read) => {
```
HEnquist, line 265: Same thought as for playbac, can `frames_read` become zero?

src/alsadevice.rs
```diff
@@ -442,57 +478,109 @@ fn playback_loop_bytes(
...
+                            timer.restart();
+                            buffer_avg.restart();
+                            if adjust {
+                                let speed = calculate_speed(
+                                    av_delay,
+                                    params.target_level,
+                                    params.adjust_period,
+                                    srate,
+                                );
+                                if let Some(elem_uac2_gadget) = &element_uac2_gadget {
```
HEnquist, line 563: We need a new setting that determines if the pitch should be adjusted on the playback or capture side. In some cases both would work, so the user should be able to choose.

src/alsadevice.rs
```diff
@@ -516,13 +604,24 @@ fn playback_loop_bytes(
...
+}
+
 fn capture_loop_bytes(
     channels: CaptureChannels,
     mut buffer: Vec<u8>,
     pcmdevice: &alsa::PCM,
     io: alsa::pcm::IO<u8>,
     params: CaptureParams,
     mut resampler: Option<Box<dyn VecResampler<PrcFmt>>>,
+    bytes_per_frame: usize,
```
HEnquist, line 624: Let's put this in the `CaptureParams` struct!

src/alsadevice.rs
```diff
@@ -609,13 +708,24 @@ fn capture_loop_bytes(
...
             }
         };
-        capture_bytes = get_nbr_capture_bytes(capture_bytes, &resampler, &params, &mut buffer);
+        // todo - rework get_nbr_capture_bytes!
+        let (new_capture_bytes, new_frames) = get_nbr_capture_bytes(capture_bytes, &resampler, &params, &mut buffer);
+        if new_capture_bytes != capture_bytes {
+            debug!("Updating capture bytes from {} to {}", capture_bytes, new_capture_bytes);
+            capture_bytes = new_capture_bytes;
+            // updating sw avail_min for snd_pcm_delay threshold
+            match update_avail_min(pcmdevice, new_frames) {
```
HEnquist, line 717: Didn't think of this! Yes definitely needed.

src/alsadevice.rs
pavhofman: I tried to find out. IMO snd_pcm_readi/writei never returns 0, but -EAGAIN. That's what my use case kept returning when the device was stalled.

src/alsadevice.rs
pavhofman: OK

src/alsadevice.rs
pavhofman: Absolutely, that patch is just a "proof of concept".

HEnquist: It looks a bit too good for being a proof of concept :) 
I don't think there's much missing before we can merge.

PR #179, opened 2022-01-23
Next100

pavhofman:
Hi Henrik,

this is fixed version of the device stall support, using the current target_level samplerate method.

### DIFF
--- a/src/alsadevice.rs
+++ b/src/alsadevice.rs
@@ -96,6 +96,7 @@ struct PlaybackParams {
     sample_format: SampleFormat,
     playback_status: Arc<RwLock<PlaybackStatus>>,
     chunksize: usize,
+    bytes_per_frame: usize,
     samplerate: usize,
 }
 
@@ -130,11 +131,12 @@ fn state_desc(state: u32) -> String {
 
 /// Play a buffer.
 fn play_buffer(
-    buffer: &[u8],
+    mut buffer: &[u8],
     pcmdevice: &alsa::PCM,
     io: &alsa::pcm::IO<u8>,
     target_delay: u64,
     millis_per_chunk: usize,
+    bytes_per_frame: usize,
 ) -> Res<()> {
     let playback_state = pcmdevice.state_raw();
     //trace!("Playback state {:?}", playback_state);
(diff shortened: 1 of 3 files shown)


src/alsadevice.rs
```diff
@@ -356,10 +398,9 @@ fn list_formats_as_text(hwp: &HwParams) -> String {
 fn open_pcm(
     devname: String,
     samplerate: u32,
-    chunksize: Frames,
-    bufsize: Frames,
     channels: u32,
     sample_format: &SampleFormat,
+    dev_buf_ctl: &mut dyn DevBufCtl,
```
HEnquist, line 403: Could we call this something less abbreviated, maybe BufferManager?

src/alsadevice.rs
```diff
@@ -681,10 +803,16 @@ fn capture_loop_bytes(
...
                 }
             }
-            Ok(CaptureResult::RecoverableError) => {
-                card_inactive = true;
-                params.capture_status.write().unwrap().state = ProcessingState::Stalled;
-                debug!("Capture device is inactive, processing is stalled");
+            Ok(CaptureResult::Stalled) => {
+                // only the first time
+                if !device_stalled {
+                    warn!("Capture device is stalled, processing is stalled");
```
HEnquist, line 809: Since this is a pretty normal condition, and not an error, should we log this on info level?

src/alsadevice.rs
```diff
@@ -951,3 +1082,123 @@ impl CaptureDevice for AlsaCaptureDevice {
         Ok(Box::new(handle))
     }
 }
+
+trait DevBufCtl {
```
HEnquist, line 1086: DeviceBufferManager?

src/alsadevice.rs
```diff
@@ -951,3 +1082,123 @@ impl CaptureDevice for AlsaCaptureDevice {
...
+    // private
+    fn _get_data(&mut self) -> &mut DevBufData;
```
HEnquist, line 1088: I see the reason for including this in the trait, but labelling it private won't make it so.. I would skip the underscore, and instead write "primarily intended for internal use" or something.

src/alsadevice.rs
```diff
@@ -951,3 +1082,123 @@ impl CaptureDevice for AlsaCaptureDevice {
...
+        rtd.bufsize - rtd.avail_min + 1
+    }
+}
+
+struct DevBufData {
+    bufsize: Frames,
+    period: Frames,
+    threshold: Frames,
+    avail_min: Frames,
+    opsize: Frames,
```
HEnquist, line 1131: opsize?

src/alsadevice.rs
```diff
@@ -951,3 +1082,123 @@ impl CaptureDevice for AlsaCaptureDevice {
...
+}
+
+struct CaptureDevBufCtl {
+    rtd: DevBufData,
```
HEnquist, line 1135: rtd? Needs a more clear name.

src/alsadevice.rs
```diff
@@ -951,3 +1082,123 @@ impl CaptureDevice for AlsaCaptureDevice {
...
+
+    fn fill_avail_min(&mut self, swp: &SwParams) -> Res<()> {
+        let rtd = self._get_data();
+        // maximum timing safety - headroom for one opsize only
+        rtd.avail_min = rtd.opsize;
+        swp.set_avail_min(rtd.opsize)?;
+        Ok(())
+    }
+
+    fn change_opsize_avail_min(&mut self, swp: &SwParams, opsize: Frames) -> Res<()> {
```
HEnquist, line 1110: Needs simpler and more clear name, isn't `update_avail_min` enough?

src/alsadevice.rs
pavhofman: OK

src/alsadevice.rs
pavhofman: There are many buffers in the chain and I wanted to emphasize that it's managing/controlling buffer of the device. Therefore I opted for the Dev (or Device) in the name. No problem with DeviceBufferManager, just quite long, especially PlaybackDeviceBufferManager :-)

src/alsadevice.rs
pavhofman: See above

src/alsadevice.rs
pavhofman: OK

src/alsadevice.rs
pavhofman: Originally it was meant as realtime data (rtd is a frequent name for structs like this). I will change rtd to data.

src/alsadevice.rs
pavhofman: It is size of read/write operation.

src/alsadevice.rs
pavhofman: It actually changes opsize and avail_min at the same time. But I will make some changes to this area today when testing shorter zero buffer to minimize latency after stall.

src/alsadevice.rs
pavhofman: Commit updated, force-pushed

src/alsadevice.rs
HEnquist: I prefer longer names if they make the code easier to read! And characters are cheap :)
I would call the trait "DeviceBufferManager", and the two structs "PlaybackBufferManager" and "CaptureBufferManager".

src/alsadevice.rs
```diff
@@ -951,3 +1082,125 @@ impl CaptureDevice for AlsaCaptureDevice {
         Ok(Box::new(handle))
     }
 }
+
+trait DevBufCtl {
+    // private
+    fn get_data(&mut self) -> &mut DevBufData;
+    fn fill_start_threshold(&mut self, swp: &SwParams) -> Res<()>;
+
+    fn fill_buffer_size(&mut self, hwp: &HwParams) -> Res<()> {
```
HEnquist, line 1091: "fill" get confusing, it makes me think of filling a buffer with data. How about `apply_buffer_size`? Same goes for the other "fill"-methods.

src/alsadevice.rs
HEnquist: `io_size`?

PR #188, opened 2022-03-02
[WIP] Add bluez-alsa audio source support

t123yh:
This adds BlueALSA audio source support.

BlueALSA ([bluez-alsa](https://github.com/Arkq/bluez-alsa)) is a project to receive or send audio through Bluetooth A2DP. In order to receive audio from it, one has to either use bluealsa-pcm ALSA plugin or kernel ALSA loopback. However, neither of these are stable: the ALSA plugin will lock up my kernel, and loopback are experiencing severe underruns (per minute). 

In order to receive audio without glitch, I added a `Bluez` capture source to read audio from the D-Bus pipe file descriptor directly. Currently, receiving audio is working.

TODO:
* Automatic detection of parameters
* Allow specifying mac address of device rather than D-Bus path
* Playback support
* Buffer level is unstable. Investigate how to get a stable buffer level.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -11,6 +11,7 @@ default = ["websocket"]
 pulse-backend = ["libpulse-simple-binding", "libpulse-binding"]
 cpal-backend = ["cpal"]
 jack-backend = ["cpal-backend", "cpal/jack"]
+bluez-backend = ["zbus"]
 32bit = []
 websocket = ["tungstenite"]
 secure-websocket = ["websocket", "native-tls", "tungstenite/native-tls"]
@@ -30,6 +31,7 @@ path = "src/bin.rs"
 alsa = "0.6.0"
 alsa-sys = "0.3.1"
 nix = "0.23"
+zbus = { version = "3.0.0", optional = true }
 
 [target.'cfg(target_os="macos")'.dependencies]
 #coreaudio-rs = { path = "../coreaudio-rs" }
--- a/README.md
+++ b/README.md
@@ -505,6 +505,19 @@ See the "camilladsp-config" repository under [Related projects](#related-project
 
 TODO test with Jack.
 
+### BlueALSA
+BlueALSA ([bluez-alsa](https://github.com/Arkq/bluez-alsa)) is a project to receive or send audio through Bluetooth A2DP. The `Bluez` source will connect to D-Bus of BlueALSA and get the audio directly from there, avoiding instability of ALSA loopback interface. Currently only capture (a2dp-sink) is supported.
+
+```
+  capture:
+    type: Bluez
+    format: S16LE
+    channels: 2
+    dbus_path: /org/bluealsa/hci0/dev_F8_87_F1_87_BF_30/a2dpsnk/source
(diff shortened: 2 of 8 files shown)


HEnquist:
This looks pretty good, let's get it into 1.1!

The way bluetooth transfers data in blocks, the buffer level will never be completely stable. The only reasonable way to ensure there are no underruns is to have a large buffer, meaning a large target level. Currently you can set it to up to 2*chunksize. Raising that limit requires some changes to the playback backends, which could be done if there are good reasons for it. But have you tried for example a chunksize of 4096 and a target level of 8192?

Auto-detecting sample rate and number of channels leads to a whole forest of new issues. There is no one-size-fits-all solution for how to modify the config in order to use the detected values. For now I would suggest reading the values and verifying that they match what the config says. The sample format can be autodetected without any issues though.

I'm not sure if it's worth spending time on bluetooth playback. I wouldn't guess that would see much use.

t123yh:
@HEnquist It's very nice to see 1.0.0 released. Could you please rebase the next11 branch onto next100, and I'll rewrite my patches on top of the next11 branch. Thank you!

HEnquist:
Sure! It will probably take a a few days until I have time to finish it though, there are some merge conflicts so I need to spend some time on it.

HEnquist:
rebase done!

HEnquist:
I looked at this one again, and noticed these things:
- The bluez support can only work on linux. Right now it's enabled for all systems, and will break compilation on anything non-linux. Please add feature flags like #[cfg(all(target_os = "linux", feature = "bluez"))], and move the dbus dependency to the linux-specific section of the manifest.
- The config reloading fix (the last commit) should really be a separate PR! And it removes support for reloading at SIGHUP while in waiting mode.

Also I'm wondering about the status of the TODO items.

HEnquist:
I finally got some time to test this, but unfortunately getting bluez-alsa to work didn't go very well.
I have paired a mobile phone to a linux laptop, and was able to play music from the phone on the laptop speakers (just worked out of the box). Then I installed bluez-alsa, and I'm stuck. Maybe because I have pipewire on the machine? Can they coexist? Disabling the pipewire service doesn't change anything.
 
`bluealsa-aplay -L` gives nothing, and `gdbus call -y --dest org.bluealsa -o /org/bluealsa -m org.freedesktop.DBus.ObjectManager.GetManagedObjects` gives the useless output `(@a{oa{sa{sv}}} {},)`

I need to be able to run this myself (with reasonable effort!) if I am to include support for it.

HEnquist:
Thanks, now it looks better. Disabling pipewire and enabling bluealsa before pairing the phone did the trick. I get useful output from both the gdbus and bluealsa-aplay commands now so I can continue.

HEnquist:
I have it working! So far it seems to work fine, I'll let it run for a while now.
The readme needs some more details, for example the need for disabling pipewire. No need to update the PR, I can take care of that later.

HEnquist:
Please see https://github.com/SmartCross/camilladsp/pull/1

PR #220, opened 2022-07-02
Updating websocket.md with string format notes

markterrill:
The server actually expects the string version of a JSON object, which it parses into JSON.

Super easy to see this from NodeJS, sending JSON.stringify("GetVersion") works. Or '"GetVersion"'. What won't work is sending "GetVersion" without wrapping in single quotations or escaping within quotes, or even worse, simply sending a javascript object that has a value like {"SetUpdateInterval": 500} 

The *simple* solution is to JSON.stringify() anything being sent by the websocket client to the CamillaDSP websocket server.

### DIFF
--- a/websocket.md
+++ b/websocket.md
@@ -8,14 +8,19 @@ By default the websocket server binds to the address 127.0.0.1, which means it's
 
 
 ## Command syntax
-All commands are sent as JSON. For commands without arguments, this is just a string with the command name within quotes:
+All commands are sent as the string text representation of a JSON object.
+
+For commands without arguments, this is just a string with the command name within quotes:
 ```
 "GetVersion"
 ```
+
 For commands that take an argument, they are instead given as a key and a value:
 ```json
 {"SetUpdateInterval": 500}
 ```
+
+Note, if you are using NodeJS/javascript then simply wrap your JSON object with JSON.stringify() before sending. IE ``` ws.send(JSON.stringify({"SetUpdateInterval": 1000}))```;
 
 The return values are also JSON. The commands that don't return a value return a structure containing the command name and the result, which is either Ok or Error:
 ```json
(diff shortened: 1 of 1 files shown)


HEnquist:
Thanks, this is a helpful addition! Just one small request, if you have the time. Could you also add a line about the reverse process of parsing the json string in a reply to an object?

markterrill:
> Thanks, this is a helpful addition! Just one small request, if you have the time. Could you also add a line about the reverse process of parsing the json string in a reply to an object?

No worries. Have done so!

markterrill:
PS, sometime I'd love for a couple new state aware functions to exist. Toggling mute, and adjusting volume proportionally. 
Background: I can now control CamillaDSP via my harmony remote control. I forked @AShifter's harmony-span (https://github.com/markterrill/harmony-span) which creates a fake Roku device on the network and added a websocket bridge to Camilla. I can now control the Camilla volume and mute from my Harmony remote, and anything that can post a webhook (even IFTTT) could then control Camilla. I'll add other functions like changing profiles/convolution filters etc.

The key thing is a remote control doesn't know the current volume and a mute button is a toggle. I sidestepped it by expanding the current calls of a simple value into an object. IE:
```json
{"SetVolume": {"changeBy": -3, "max": 30, "min":-35}}
{"SetMute":"toggle"}
```
The harmony-span checks the current button function, if it detects it's the 'expanded' version of those commands it first requests current volume / current mute state, then issues the appropriate set in concrete value.

Looking at camilla code, while not familiar with the language it looks fairly easy to modify. What I don't know is if you can have overloaded functions with different values for the arguments. I'd need help with that. Otherwise, the fallback would be new function names, like ```ChangeVolumeBy(value, min, max)``` and ToggleMute().
What are your thoughts?

HEnquist:
Merged, thanks!
Sorry about the late reply. For some reason I didn't get any notification for the last message.
I can definitely add the commands you mention. I would add them as completely new commands ToggleMute, and AdjustVolume(+- change in dB). Both should probably return the new value in the reply. I would prefer to have a single AdjustVolume that can take both positive and negative values, instead of having separate commands for raising and lowering.

Issue #239, opened 2022-10-13
cdsp doesn't stay muted after switching conf files (yet reports being muted)


taradiddles:
minor bug:

in camillaGui (or via pycamilladsp's api), with a working config / some music playing:

- mute volume (mute icon above volume bars or set_mute())
- in the files tab: load another config in the gui
- 'apply to dsp'

-> the mute icon stays toggled (or get_mute() returns True) but the music/signal is played at normal volume. To (really) mute I have to un-toggle and re-toggle.

thanks for your work !

[edit: filed in this repo as it seems the issue has something to do with camilladsp's internals, but it could just be an issue in how pycamilladsp exposes the mute property]

[edit2: changing the volume while camilladsp is muted but playing at normal volume (just after the switch to the new config file) actually mutes the output]

HEnquist:
Thanks for reporting! I took a quick look at the code and I think I know what is wrong. I'll fix it and release a new version asap.

HEnquist:
Fixed in 1.0.3

PR #242, opened 2022-10-25
Respect mute state on reload

HEnquist:


HEnquist:
Fixes https://github.com/HEnquist/camilladsp/issues/239

PR #247, opened 2022-12-06
Dither improvements

roderickvd:
This PR contains various improvements to the "Dither" filter. For ease of reviewing, they are split up in separate commits. I can squash if so desired.

Some accompanying notes and questions:

**Improved CPU load**
Switched to the [SmallRng](https://docs.rs/rand/latest/rand/rngs/struct.SmallRng.html) PRNG. This consistently shaves off about one percent point on my Core i5 iMac, even without more extensive profiling.

Another percentage point is shaved off by changing the two-tap `[1, 0.8]` "Simple" ditherer to a Wannamaker `[1, -1]` high pass. This only requires one PRNG sample per audio sample, and still yields a triangular distribution by combining it with the previous PRNG sample. I thought this was OK to change in-place as it still yields a high passed / violet-shaped dither.

**Additional ditherers**
Added some well-known variants of existing ditherers. An endless number of noise shaping curves can be found when scouring the net, but these seem a concise selection of well-regarded, well-known ones in the audio community.

In particular, almost everyone seems to be using the old Shibata filters, probably unaware that there is a new set of filters. The new set has filters that are smoother, and also support 88.2, 96 and 192 kHz.

This requires some supporting changes to https://github.com/HEnquist/camillagui and https://github.com/HEnquist/pycamilladsp-plot which I will push shortly.

**Questions**
- As I create this PR I see a [`next20`](https://github.com/HEnquist/camilladsp/commits/next20) branch emerging. Do you want me to rebase on that?
- The "Uniform" ditherer is suboptimal sound quality-wise. While it does offer slightly lower cpu load than "Triangular", it is the same as "Simple" so might as well use that instead. Could "Uniform" be removed?
- The "None" ditherer seems an edge-case (why not just remove the "Dither" filter from the pipeline?) and breaks audio output on low-bit settings due to scaling and rounding. This is also how it was before this PR, and seems counter-intuitive. Could it be removed?
- "Simple" seems imprecise about what it does. Would "HighPass" be a better name?

**Possible future work**
- The noise shapers are hard-coded to use the triangular pdf. This PR enables under-the-hood support to have the noise shapers use any pdf. Some other software packages have separate drop-downs for the pdf and noise shaper. The CamillaDSP API and GUI could be made that way too. Although objectively, there is no need to use anything but triangular.
- To cut down on the number of subtypes, an extra drop-down could be added to select from a "family" of noise shapers. For example: instead of having Shibata441, Shibata48, ShibataLow441, and so on, you could select "Family: Shibata" and then a third drop-down "Subtype: ShibataLow441".

Let me know if you have interest in any of that, and I can work on it.


HEnquist:
> This PR contains various improvements to the "Dither" filter. For ease of reviewing, they are split up in separate commits. I can squash if so desired.
> 

Thanks for submitting this PR! Keeping the commits is fine, need to squash.

> 
> **Improved CPU load** Switched to the [SmallRng](https://docs.rs/rand/latest/rand/rngs/struct.SmallRng.html) PRNG. This consistently shaves off about one percent point on my Core i5 iMac, even without more extensive profiling.
>

Extra speed is always welcome! 

> Another percentage point is shaved off by changing the two-tap `[1, 0.8]` "Simple" ditherer to a Wannamaker `[1, -1]` high pass. This only requires one PRNG sample per audio sample, and still yields a triangular distribution by combining it with the previous PRNG sample. I thought this was OK to change in-place as it still yields a high passed / violet-shaped dither.

Same as above  :)


> 
> **Additional ditherers** Added some well-known variants of existing ditherers. An endless number of noise shaping curves can be found when scouring the net, but these seem a concise selection of well-regarded, well-known ones in the audio community.
> 
> In particular, almost everyone seems to be using the old Shibata filters, probably unaware that there is a new set of filters. The new set has filters that are smoother, and also support 88.2, 96 and 192 kHz.

Yes there are quite a few out there. The existing set is basically a quick collection of the first ones I found. Extending the Shibata set to more sample rates is certainly useful.
 
> 
> This requires some supporting changes to https://github.com/HEnquist/camillagui and https://github.com/HEnquist/pycamilladsp-plot which I will push shortly.
> 
> **Questions**
> 
> * As I create this PR I see a [`next20`](https://github.com/HEnquist/camilladsp/commits/next20) branch emerging. Do you want me to rebase on that?

Yes that would be great. I needed to make breaking changes to the config file format, so I decided to skip v1.1 and go directly for 2.0. It would save time later on if you rebase on next20 (should be quite easy I think) and change the PR to target that one.

> * The "Uniform" ditherer is suboptimal sound quality-wise. While it does offer slightly lower cpu load than "Triangular", it is the same as "Simple" so might as well use that instead. Could "Uniform" be removed?
> * The "None" ditherer seems an edge-case (why not just remove the "Dither" filter from the pipeline?) and breaks audio output on low-bit settings due to scaling and rounding. This is also how it was before this PR, and seems counter-intuitive. Could it be removed?

The Uniform and None aren't very useful for actual playback, but that's not their purpose. The idea is to provide a toy box of different dithering strategies so people can experiment. If you use None, and set the bit depth very low (4-5 bits for example) then it sounds pretty awful, especially on simple material like a single voice. Then switch to Uniform with the same bit depth. Much better, but noisy. Then from there you can go on and try the different noise shapers to learn how they sound. IMO much better and certainly more fun than just picking a cryptic name that you read somewhere is supposed to be good.

The name "Uniform" refers to that the noise spectrum is uniform. It should perhaps be renamed to "Unshaped". 
I haven't tried them yet but I suppose the new Uniform/Triangular/Gaussian perform similar to the old Uniform (meaning mostly good for toying with). Wouldn't a single "Unshaped" be sufficient here?

> * "Simple" seems imprecise about what it does. Would "HighPass" be a better name?

Yes, seems like a better name.

> 
> **Possible future work**
> 
> * The noise shapers are hard-coded to use the triangular pdf. This PR enables under-the-hood support to have the noise shapers use any pdf. Some other software packages have separate drop-downs for the pdf and noise shaper. The CamillaDSP API and GUI could be made that way too. Although objectively, there is no need to use anything but triangular.

If there is no need to ever use anything but triangular, then I think it's best to only offer that. Too many options just gets overwhelming.

> * To cut down on the number of subtypes, an extra drop-down could be added to select from a "family" of noise shapers. For example: instead of having Shibata441, Shibata48, ShibataLow441, and so on, you could select "Family: Shibata" and then a third drop-down "Subtype: ShibataLow441".

This could be an option, but I think the list of shapers is still short enough to be ok in a single list. If it turns out to be too long, there is also the option of removing some shapers. The Fweighted for example, does it have any advantages compared to the fancier ones?

> 
> Let me know if you have interest in any of that, and I can work on it.

roderickvd:
> Thanks for submitting this PR! Keeping the commits is fine, need to squash.

Happy to help! I love Rust and working on audio software, so looking forward to help where I can.

> > * As I create this PR I see a [`next20`](https://github.com/HEnquist/camilladsp/commits/next20) branch emerging. Do you want me to rebase on that?
> 
> Yes that would be great. I needed to make breaking changes to the config file format, so I decided to skip v1.1 and go directly for 2.0. It would save time later on if you rebase on next20 (should be quite easy I think) and change the PR to target that one.

Will do.

> The Uniform and None aren't very useful for actual playback, but that's not their purpose. The idea is to provide a toy box of different dithering strategies so people can experiment.

OK, got it.

> If you use None, and set the bit depth very low (4-5 bits for example) then it sounds pretty awful, especially on simple material like a single voice.

Understood *however* I think that's currently broken on "None" because of rounding, then scaling. On many playback volumes below 0 dB it is silent (I guess all rounded down to 0, did not check extensively), only when you turn it up it starts outputting 1's and immediately clips.

We could fix this by skipping through `process_waveform()` entirely, which would involve some extra boilerplate code to check whether the filter is `None` or `Some`.

> The name "Uniform" refers to that the noise spectrum is uniform. It should perhaps be renamed to "Unshaped". I haven't tried them yet but I suppose the new Uniform/Triangular/Gaussian perform similar to the old Uniform (meaning mostly good for toying with). Wouldn't a single "Unshaped" be sufficient here?

May I suggest "Flat"?

In this PR all three pdfs are flat: Triangular, Uniform and Gaussian.
The pre-PR "Uniform" is now "Triangular".
Continuing below...

> > * The noise shapers are hard-coded to use the triangular pdf. This PR enables under-the-hood support to have the noise shapers use any pdf. Some other software packages have separate drop-downs for the pdf and noise shaper. The CamillaDSP API and GUI could be made that way too. Although objectively, there is no need to use anything but triangular.
> 
> If there is no need to ever use anything but triangular, then I think it's best to only offer that. Too many options just gets overwhelming.

...I would then suggest to remove Uniform and Gaussian.
Keeping only Triangular and call that "Flat".
What do you think?

> > * To cut down on the number of subtypes, an extra drop-down could be added to select from a "family" of noise shapers. For example: instead of having Shibata441, Shibata48, ShibataLow441, and so on, you could select "Family: Shibata" and then a third drop-down "Subtype: ShibataLow441".
> 
> This could be an option, but I think the list of shapers is still short enough to be ok in a single list. If it turns out to be too long, there is also the option of removing some shapers. The Fweighted for example, does it have any advantages compared to the fancier ones?

Well, one thing I learned when working on [librespot](https://github.com/librespot-org/librespot) is: once you offer something, you really don't know what people are using it for.

The Fweighted family is based on an early version of what is now ISO 226. The Shibata filters are based on the LAME ATH curves (Absolute Threshold of Hearing), which are in turn based on the Spanias & Painter ATH formula.

If anything you could choose to drop the Lipshitz filters, which are not corrected for diffuse-field listening, but then someone will come along and say "hey where did my filters for open air concerts go?" 😆

HEnquist:
> Understood _however_ I think that's currently broken on "None" because of rounding, then scaling. On many playback volumes below 0 dB it is silent (I guess all rounded down to 0, did not check extensively), only when you turn it up it starts outputting 1's and immediately clips.

Yes with very few bits it tends to go quiet, when there isn't enough signal to reach even one LSB. I would argue that this is the correct behaviour.

> 
> We could fix this by skipping through `process_waveform()` entirely, which would involve some extra boilerplate code to check whether the filter is `None` or `Some`.

Maybe I'm missing something, but then we wouldn't get the desired bit depth reduction..?  

> 
> ...I would then suggest to remove Uniform and Gaussian. Keeping only Triangular and call that "Flat". What do you think?
> 
Perfect!


> Well, one thing I learned when working on [librespot](https://github.com/librespot-org/librespot) is: once you offer something, you really don't know what people are using it for.

Interesting, I also worked a bit on librespot :) But it was a while ago now, when I just had discovered how much fun Rust is.

> If anything you could choose to drop the Lipshitz filters, which are not corrected for diffuse-field listening, but then someone will come along and say "hey where did my filters for open air concerts go?" 😆

:D 
But let's keep poor old Lipshitz for now, it's easy to kick him out later if needed.

src/dither.rs
```diff
@@ -1,260 +1,761 @@
...
+            -0.018_070_342_019_200_325,
+            -0.013_997_578_062_117_1,
+        ])
+    }
+
+    pub fn process(&mut self, scaled: PrcFmt, dither: PrcFmt) -> PrcFmt {
+        let filterlen = self.filter.len();
+        let mut filt_buf = 0.0;
+        for (n, coeff) in self.filter.iter().enumerate() {
+            filt_buf += coeff * self.buffer[(n + self.idx) % filterlen];
```
HEnquist, line 451: I think `self.buffer` could be changed to a VecDeque to simplify things here. Just pushing and popping samples at eather end, and the VecDeque handles the indexing. May also be faster.

HEnquist: This can also be done later, just a suggestion if you think it's interesting.
(And if you think it's really interesting, there is also a homemade `FifoQueue` in the Delay filter. That could likely also use VecDeque instead and perhaps get a little faster)

CHANGELOG.md
```diff
@@ -14,6 +14,18 @@ New features:
 - Add title and description fields to various parts of the config.
 - Gain can be specified in dB or linear scale.
 - Websocket command to reset clipped samples counter.
+- Add ditherers:
```
HEnquist, line 17: This gets a little long, but let's leave it like this. I will compress the changelog once it's getting closer to a release.

README.md
```diff
@@ -1451,32 +1451,53 @@ Other types such as Bessel filters can be built by combining several Biquads. [S
...
+| Shibata48           | 48   | 16 taps       | Modeled after LAME ATH type 1                                  |
+| - ShibataHigh48     | 48   | 28 taps       | - High intensity (quieter noise)                               |
+| - ShibataLow48      | 48   | 16 taps       | - Low intensity (louder noise)                                 |
+| Shibata882          | 88.2 | 20 taps       | Modeled after LAME ATH type 1                                  |
+| - ShibataLow882     | 88.2 | 24 taps       | - Low intensity (louder noise)                                 |
+| Shibata96           | 96   | 31 taps       | Modeled after LAME ATH type 1                                  |
+| - ShibataLow96      | 96   | 32 taps       | - Low intensity (louder noise)                                 |
+| Shibata192          | 192  | 54 taps       | Modeled after LAME ATH type 1                                  |
+| - ShibataLow192     | 192  | 20 taps       | - Low intensity (louder noise)                                 |
+
```
HEnquist, line 1480: Perhaps add a something about more taps -> higher cpu usage?

HEnquist:
I went through this properly now, looks very good! I made just a few minor comments.

src/dither.rs
roderickvd: Good idea. I'll get the `VecDeque` in here and tackle it in the `Delay` filter later.

README.md
roderickvd: Will do.

roderickvd: Done in https://github.com/HEnquist/camilladsp/pull/247/commits/b383ced4523b2d1fe897efedd264441961a96009.

src/dither.rs
roderickvd: Done in https://github.com/HEnquist/camilladsp/pull/247/commits/75e691087d60207cc817e82f9e924160b18600a5.

I ended up using the `ringbuffer` crate instead of `VecDeque`. The latter is slower than a normal `Vec` for iterations, which is the main operation in need of optimisation, particularly in the case of many taps. Also it's growable, which is OK, but we really just need something with a fixed size.

I considered several ring buffer crates. I chose this one because it seems well-maintained, more than enough downloads, uses idiomatic Rust (i.e. implements a `DoubleEndedIterator` to make a LIFO queue), and is fast.

In the end it does not seem like it's any faster (might be in micro-benchmarks) and pulls in an extra dependency. But this does improve readability, and no doubt will be of use in other parts of the source code as well.

roderickvd:
I addressed your review comments and rebased. Please see https://github.com/HEnquist/camilladsp/pull/247#discussion_r1045258343 for some discussion on `VecDeque`.

HEnquist:
> I addressed your review comments and rebased. Please see [#247 (comment)](https://github.com/HEnquist/camilladsp/pull/247#discussion_r1045258343) for some discussion on `VecDeque`.

Great! I saw that ring buffer got added and I was going to ask why, but then I read that comment and it makes perfect sense.

PR #253, opened 2022-12-12
More idiomatic getter names and string slices

roderickvd:
As discussed before. This basically touches every file :smile:
While there it was also low-effort to pre-allocate `Vec`s when we know the size up front.

### DIFF
--- a/src/diffeq.rs
+++ b/src/diffeq.rs
@@ -20,7 +20,9 @@ pub struct DiffEq {
 }
 
 impl DiffEq {
-    pub fn new(name: String, a_in: Vec<PrcFmt>, b_in: Vec<PrcFmt>) -> Self {
+    pub fn new(name: &str, a_in: Vec<PrcFmt>, b_in: Vec<PrcFmt>) -> Self {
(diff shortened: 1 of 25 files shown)


src/diffeq.rs
```diff
@@ -43,9 +45,9 @@ impl DiffEq {
         }
     }
 
-    pub fn from_config(name: String, conf: config::DiffEqParameters) -> Self {
-        let a = conf.get_a();
-        let b = conf.get_b();
+    pub fn from_config(name: &str, conf: config::DiffEqParameters) -> Self {
+        let a = conf.a.unwrap_or_default();
```
HEnquist, line 49: These should use the getters for a and b.

HEnquist:
Looks good! Just one small comment about some unused getters.

src/diffeq.rs
roderickvd: Good catch. Done in https://github.com/HEnquist/camilladsp/pull/253/commits/56c62382dc2bb7d30c72248d63b934920142d6c5.

PR #254, opened 2022-12-14
Optimize delay filter

roderickvd:
Replacing `FifoQueue` with `CircularQueue` nicely speeds up the delay filter in the order of 20-25% (relative) on my headless `aarch64` gadget.

In hindsight I should have done this as part of https://github.com/HEnquist/camilladsp/pull/247. The `RingBuffer` I used there is fast but requires a capacity that is a power of two. That does not work well with a delay filter of course. I now replaced `RingBuffer` which is more generic, but probably not as fast, but faster than what we had.

While there I took the opportunity to refactor the delay filter. Some points to check for subsample precision:
- I don't believe the sample rate passed to the `Biquad` matters, but I changed it to the actual rate instead of 12345.
- Normally, mathematically, a fraction is calculated as the difference to a truncated value, not a floored value. It should only matter for negative values, which are forbidden here, but still.


roderickvd:
That's weird. A failing noise shaper only on Windows? The CI for this commit ran fine on my repo: https://github.com/roderickvd/camilladsp/actions/runs/3698619244

HEnquist:
The dither adds randomness, this probably means that the test doesn't take that into account properly. I have never seen it before so it seems very rare, let's ignore for now. Just running it again will probably make it pass.

roderickvd:
Yeah all good now, thanks. I'll take a look at that later if and when it surfaces again.

I edited the PR subject and description where I was mixing up the dither and delay filters. This PR is about the delay filter primarily. Time for bed I guess 😴

src/basicfilters.rs
```diff
@@ -308,13 +311,14 @@ impl Filter for Gain {
...
 impl Delay {
     /// Creates a delay filter with delay in samples
-    /// Will be improved as it gets slow for long delays
-    pub fn new(name: String, samplerate: usize, delay: PrcFmt, subsample: bool) -> Self {
+    pub fn new(name: &str, samplerate: usize, delay: PrcFmt, subsample: bool) -> Self {
+        let name = name.to_string();
+
         let (integerdelay, biquad) = if subsample {
-            let samples = delay.floor();
+            let samples = delay.trunc();
```
HEnquist, line 318: This was using floor(), because the fractional delay must be positive. It doesn't matter in practice since negative delays aren't a thing, but I would still prefer to have floor here.
I have no idea what a biquad created with a negative `fraction` would do. Probably nothing good..

HEnquist:
> * I don't believe the sample rate passed to the `Biquad` matters, but I changed it to the actual rate instead of 12345.

Yes the 12345 was just a placeholder. The actual values doesn't matter, but of course having the true one doesn't hurt.

src/basicfilters.rs
roderickvd: Reverted in https://github.com/HEnquist/camilladsp/pull/254/commits/71038a4c398cf7176c4f4b19f6de87409b48ab41.

roderickvd:
Rebased on the latest `next20` and addressed your review comment.

PR #255, opened 2022-12-15
Fix compiling with fftw

roderickvd:
No longer compiled because of https://github.com/HEnquist/camilladsp/pull/253 and https://github.com/HEnquist/camilladsp/commit/5d9bb1fd8d02208ec64fecd85e80b9c449eef6da.

Maybe add a CI action with all the optionals turned on? Can add it if you'd like that.

### DIFF
--- a/src/fftconv_fftw.rs
+++ b/src/fftconv_fftw.rs
@@ -102,7 +102,8 @@ pub struct FftConv {
 
 impl FftConv {
     /// Create a new FFT colvolution filter.
-    pub fn new(name: String, data_length: usize, coeffs: &[PrcFmt]) -> Self {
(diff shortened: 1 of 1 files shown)


HEnquist:
I haven't decided what to do with FFTW yet. RustFFT has gotten quite a bit faster since I added this option, so I'm not sure if it's worth keeping the FFTW option.

HEnquist:
Regardless of what happens with FFTW, a CI run with all optionals turned on would be really useful.

HEnquist:
Merged. Since there is no additional work to keep it now, it can stay. If maintaining it becomes a burden at some point, then that means it's time to remove.

PR #272, opened 2023-03-14
Add metadata field to config

simonjohansson:
Hello! Tiny PR to allow a metadata field in the config. For me this would be useful in two ways.

## Anchors. 

Im numerically challenged so I cannot remember what output 5 is without a cheat sheet. So by putting anchors at the top in the metadata it makes it really easy to change, and reason about the config further down.

```yaml
---
metadata: 
  channels:
    inL: &inL 0
    inR: &inR 1
    monitorL: &monitorL 0
    monitorR: &monitorR 1
    headphoneL: &headphoneL 2
    headphoneR: &headphoneR 3
    sub1: &sub1 4
    sub2: &sub2 5
  ....
....
mixers:
  stereo:
    channels:
      in: 2
      out: 10
    mapping:
    - dest: *headphoneL
      sources:
      - channel: *inL
    - dest: *headphoneR
      sources:
      - channel: *inR
     ...
...
```

We could ofc define the anchors directly in the mixers and re-use them further down, but I like the neatness of having it at the top.

## Putting arbitrary metadata in the config

Im building a physical control surface for my CamillaDSP setup, to allow me to quickly switch on/off headphones/monitors, disable sub1, sub2, change room correction depending on if my standup table is down or up, switches for different crossfeed configs for the headphones etc etc. The permutations of different configs has exploded, if I could add arbitrary metadata I could parse that from a ESP32 or similar and set led on/off, make sure the control surface is up to date with the config etc.

Given a active CamillaDSP config with metadata

```yaml
---
metadata: 
    ...
  enabled:
    monitors: false
    headphones: true
```

I could easily poll over websocket and do something like

```python
while True:
    current_meta = cdsp.get_config()['metadata']
    toggleLED(MONITORS_ENABLED_LED, current_meta['enabled']['monitors'])
    toggleLED(HEADPHONE_ENABLED_LED, current_meta['enabled']['headphones'])
```

Hope that makes sense.

### DIFF
--- a/src/config.rs
+++ b/src/config.rs
@@ -1,6 +1,7 @@
 use crate::filters;
 use crate::mixer;
 use serde::{de, Deserialize, Serialize};
+use serde_json::{Value};
 use serde_with;
 use std::collections::HashMap;
 use std::error;
@@ -789,6 +790,8 @@ pub struct Configuration {
     pub filters: HashMap<String, Filter>,
     #[serde(default)]
     pub pipeline: Vec<PipelineStep>,
+    #[serde(default)]
+    pub metadata: Value,
 }
 
 fn validate_nonzero_usize<'de, D>(d: D) -> Result<usize, D::Error>


HEnquist:
Neat with the anchors! I haven't used that at all, mainly because they don't survive a deserialization/serialization roundtrip. But as long as config files are only edited manually it's fine of course.

I fear that the metadata field would see very little use. The completely free content also means it's not really feasible to show and edit it in the gui. The next version of camilladsp (in the next20 branch) has a "description" field in the config for text. Could you instead use that to store a json object as a string?

PR #285, opened 2023-05-28
Add websocket commands for listing devices

HEnquist:


Wang-Yue:
this is awesome!

It would be even more useful if this command could list all supported formats (sample rate and bit depth) so a GUI program can rely on this information to do a drop down menu for selection.

HEnquist:
There will be separate commands for listing stream configs for a device. But I'll push that to after 2.0 is released. It's done quite differently in different audio apis, so won't be as straightforward as just listing the devices.

Wang-Yue:
yeah that makes perfect sense. Thanks!

PR #286, opened 2023-06-02
reset StopReason when PlaybackReady

Wang-Yue:
Lemme know if I misunderstand the code but IMHO CaptureReady and PlaybackReady should be treated in the same way. When the loop starts the stopreason should be reset.
I don't really see the past code causing problem in reality, most probably because 1) the value is initialized to none 2) playback thread are always running before the capture thread, and by the time capture is ready, the stop reason is reset there.

### DIFF
--- a/src/bin.rs
+++ b/src/bin.rs
@@ -327,6 +327,7 @@ fn run(
                         barrier.wait();
                         debug!("Supervisor loop starts now!");
                         is_starting = false;
+                        status_structs.status.write().unwrap().stop_reason = StopReason::None;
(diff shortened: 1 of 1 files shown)


HEnquist:
Thanks, you are right that it should be reset the same way for capture and playback. But looking at it now I don't think it should be reset here at all. It is probably better to do it before that main loop starts.
I'll take a proper look and make sure it's handled correctly in v2.0. It's probably harmless enough that we can skip fixing it in a 1.0.x release.

PR #287, opened 2023-06-18
[RFC] Improving config reload handling

t123yh:
In my application there's a case where two config change message will be sent through WebSocket in a very short time. Sometimes, the second config won't take effect. We found that sending reload signal using an atomic boolean and a poll timer is not a reliable way to pass message, and may lead to race condition.

The main idea of this pull request is to remove the use of shared variable to pass reload and stop signal, and replace it with a crossbeam channel (control message channel). In the main loop, instead of polling the atomic variable with a fixed interval controlled by queue timeout, we use crossbeam::select! to simultaneously receive from the control channel and status channel, and respond to whatever event instantly (instead of having to wait for at most 100ms to respond to reload request).

When config is being changed by signal or websocket, new config is no longer passed by the `new_config` globally shared variable. Instead, it is passed through the control channel. This ensures no race condition will occur. Signals are now handled by a separate polling thread. When the signal handling thread receives a signal, it will respond by sending message to the control channel. In addition to SIGHUP, I also implemented SIGUSR1, which sends a Stop message to control thread.

I've tested the control logic by signal and websocket, and it works fine. However I'm sure I haven't covered all use cases, and it probably requires more testing. I don't have a Windows environment, so Windows support is untested.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -38,14 +38,12 @@ zbus = { version = "3.0.0", optional = true }
 #coreaudio-rs = { path = "../coreaudio-rs" }
 #coreaudio-rs = { git = "https://github.com/HEnquist/coreaudio-rs", tag="v0.11.1-rc1" }
 coreaudio-rs = "0.11.1"
-crossbeam-channel = "0.5"
 dispatch = "0.2.0"
 
 [target.'cfg(target_os="windows")'.dependencies]
 #wasapi = { path = "../../rust/wasapi" }
 #wasapi = { git = "https://github.com/HEnquist/wasapi-rs", branch = "win041" }
 wasapi = "0.13.0"
-crossbeam-channel = "0.5"
 windows = {version = "0.48.0", features = ["Win32_System_Threading", "Win32_Foundation"] }
 
 [dependencies]
@@ -78,6 +76,7 @@ cpal = { version = "0.13.3", optional = true }
 rawsample = "0.2.0"
 circular-queue = "0.2.6"
 parking_lot = { version = "0.12.1", features = ["hardware-lock-elision"] }
+crossbeam-channel = "0.5"
 
 [build-dependencies]
 version_check = "0.9"
--- a/src/alsadevice.rs
+++ b/src/alsadevice.rs
@@ -66,13 +66,13 @@ pub struct AlsaCaptureDevice {
 
 struct CaptureChannels {
     audio: mpsc::SyncSender<AudioMessage>,
-    status: mpsc::Sender<StatusMessage>,
+    status: crossbeam_channel::Sender<StatusMessage>,
     command: mpsc::Receiver<CommandMessage>,
 }
 
 struct PlaybackChannels {
     audio: mpsc::Receiver<AudioMessage>,
-    status: mpsc::Sender<StatusMessage>,
+    status: crossbeam_channel::Sender<StatusMessage>,
 }
 
 struct CaptureParams {
@@ -908,7 +908,7 @@ impl PlaybackDevice for AlsaPlaybackDevice {
         &mut self,
         channel: mpsc::Receiver<AudioMessage>,
         barrier: Arc<Barrier>,
-        status_channel: mpsc::Sender<StatusMessage>,
+        status_channel: crossbeam_channel::Sender<StatusMessage>,
         playback_status: Arc<RwLock<PlaybackStatus>>,
     ) -> Res<Box<thread::JoinHandle<()>>> {
         let devname = self.devname.clone();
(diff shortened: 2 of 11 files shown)


HEnquist:
I started thinking a bit about the manual testing procedure. The first thing is definitely to define the desired behaviour.
I'm thinking of what happens when several reload requests are received (via websocket) faster than they can be applied. Some things to think about:
- Should a reload request be accepted while another is being processed? (if not, the client must of course be notified that the request was denied)
- If yes, what should be the limit for number of queued requests? Unlimited is too risky, some sane limit is needed.
- For a quick burst of requests that arrive while another one is being processed, should they all be applied, or only the last one? Applying all may take some time if they are many. Debouncing by skipping the intermediate ones would avoid wasting resources on applying something that will immediately be replaced.

t123yh:
We should accept reload requests, in any time.

Why do you say unlimited is risky? I don't get what risk means here.

For a burst of requests, I think the last one should be processed. So even if there are many requests queued, only the last one will be processed (unless there's EXIT request in the middle). 

The logic is that, we try to make sure that the currently used config match with the latest config submitted by user.

HEnquist:
> We should accept reload requests, in any time.

Why? Not saying I disagree, just looking for a motivation. One could also argue that since reload requests may be quite expensive to process, they should be carefully throttled by the client.

> 
> Why do you say unlimited is risky? I don't get what risk means here.

Unlimited is risky because then there is no limit for how much memory the queue can consume. It's allowed to put large sets of fir filter coefficients directly in the config file, and then each request takes some space. Probably not something that will cause trouble very often, but it's still better to put a reasonable limit for the size. I would say that any client that sends so many requests (and so quickly) that more than a handful end up queued is severely flawed. I would start with a limit of 10 or so. 

> 
> For a burst of requests, I think the last one should be processed. So even if there are many requests queued, only the last one will be processed (unless there's EXIT request in the middle).
> 
> The logic is that, we try to make sure that the currently used config match with the latest config submitted by user.

Definitely it must be guaranteed that the last submitted request that was accepted is the active one. We have to do better than "try to make sure" :)

t123yh:
> > We should accept reload requests, in any time.
> 
> Why? Not saying I disagree, just looking for a motivation. One could also argue that since reload requests may be quite expensive to process, they should be carefully throttled by the client.
In my use case, a daemon monitors several input sources, and switch input source based on priority. Some source status may change relative quickly. If we just reject reload request while starting, some retry logic must be implemented in the daemon, which I think is not desirable.

Although "reload requests may be quite expensive to process", when several reload requests are queued, we should just drop the requests in the middle and directly process the request in the end. This can avoid unnecessary processing time. The only thing that matters here is the memory pressure caused by "large sets of fir filter coefficients" in the command queue. 

> 
> > Why do you say unlimited is risky? I don't get what risk means here.
> 
> Unlimited is risky because then there is no limit for how much memory the queue can consume. It's allowed to put large sets of fir filter coefficients directly in the config file, and then each request takes some space. Probably not something that will cause trouble very often, but it's still better to put a reasonable limit for the size. I would say that any client that sends so many requests (and so quickly) that more than a handful end up queued is severely flawed. I would start with a limit of 10 or so.

Ok, I agree that a limit is needed.

> 
> > For a burst of requests, I think the last one should be processed. So even if there are many requests queued, only the last one will be processed (unless there's EXIT request in the middle).
> > The logic is that, we try to make sure that the currently used config match with the latest config submitted by user.
> 
> Definitely it must be guaranteed that the last submitted request that was accepted is the active one. We have to do better than "try to make sure" :)

Yes.

HEnquist:
Thanks for explaining your reasoning. Let's go with this approach of accepting all reload requests (until hitting the queue limit), and then dropping intermediate requests.

HEnquist:
I started on a simple testing script that applies a few configs and checks that the correct one becomes active. The script is here: https://github.com/HEnquist/camilladsp/commit/153dea1259b111e591f0bfc05c38ce8dfe6fd1d4

It starts by changing very slowly, then with a 100ms delay, and finally with a 1 ms delay. The current version fails after a few iterations in the 100ms test. But when running this PR, it fails on the first assert because it gets None for the active config. It seems like the active config isn't returned correctly. Can you check?

t123yh:
> I started on a simple testing script that applies a few configs and checks that the correct one becomes active. The script is here: [153dea1](https://github.com/HEnquist/camilladsp/commit/153dea1259b111e591f0bfc05c38ce8dfe6fd1d4)
> 
> It starts by changing very slowly, then with a 100ms delay, and finally with a 1 ms delay. The current version fails after a few iterations in the 100ms test. But when running this PR, it fails on the first assert because it gets None for the active config. It seems like the active config isn't returned correctly. Can you check?

I fixed this error, and also rebased my commits to latest next20, could you please try again?

HEnquist:
Yes now it works great, thanks!
I updated the test script to use purest, and to test more ways of changing the config. Using your branch, all tests pass, while next20 fails on a bunch.
I think the only things left here is to drop intermediate reload requests, and to change that unlimited channel to a limited one. In the unlikely event that the channel gets full, it would be nice to start dropping the oldest not-yet-applied requests instead of refusing the request. But it's possible that this is so unlikely (especially if the intermediate ones are dropped) that we should not spend time on it.

HEnquist:
Merged!
I implemented the changes that were left in https://github.com/HEnquist/camilladsp/pull/300
I also simplified the wait loop a bit. @t123yh please take a look.

PR #291, opened 2023-07-27
Compressor: Fix correct clip limit to Limiter

ElitCoder:
The input to Limiter should be from the configuration, otherwise the Limiter will try to use the linear gain as the dB clip limit.


HEnquist:
Good job spotting this one :) 
Thanks for the PR!

PR #299, opened 2023-08-05
add volumio reference

balbuze:


HEnquist:
Thanks! Just a question about the Google Docs link. If you update the docs, will the new version be published on the same url?

balbuze:
hello! yes, this is the link for the online help in the plugin. It is updated automatically when I change the source.
image

PR #302, opened 2023-10-08
Add workaround for rustc bug https://github.com/rust-lang/rust/issues…

spfenwick:
…/116359

by zeroing all sse2 registers on x86-64 architectures at start of processing loop

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "camilladsp"
-version = "2.0.0-alpha3"
+version = "2.0.0-alpha4"
 authors = ["Henrik Enquist <henrik.enquist@gmail.com>"]
 edition = "2021"
 description = "A flexible tool for processing audio"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -17,6 +17,7 @@ websocket = ["tungstenite"]
 secure-websocket = ["websocket", "native-tls", "tungstenite/native-tls"]
 FFTW = ["fftw"]
 debug = []
+avoid-rustc-issue-116359 = []
 
 [lib]
 name = "camillalib"
--- a/src/filters.rs
+++ b/src/filters.rs
@@ -434,7 +434,7 @@ impl FilterGroup {
         if !input.waveforms[self.channel].is_empty() {
             // Zeroes all sse registers on x86_64 architecturesto work around
             // rustc bug https://github.com/rust-lang/rust/issues/116359
-            #[cfg(target_arch = "x86_64")]
+            #[cfg(all(target_arch = "x86_64", feature="avoid-rustc-issue-116359"))]
             unsafe {
                 use std::arch::asm;
                 asm!(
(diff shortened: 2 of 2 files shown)


spfenwick:
As per https://www.diyaudio.com/community/threads/camilladsp-cross-platform-iir-and-fir-engine-for-crossovers-room-correction-etc.349818/post-7474900 it looks like setting codegen_units = 1 has not fully avoided the denormal performance issues. This workaround provides a fix in all the scenarios I've tested.

HEnquist:
Thanks for the PR! What do you think about making the workaround optional via a cargo feature? I think that would be a bit nicer than including it by default.

HEnquist:
I have looked at this and I see a real risk that the problem can still be triggered by some config files. This only clears the sse registers at the start of the processing. If any filter leaves something behind there before a biquad, then this won't help. There are many combinations of filters to test to make sure.
I think it's safer to bump the cpu requirements with `target-cpu=x86-64-v2`. That compiles into something quite different that doesn't have the problem at all.

This woraround would still be useful for compiling for the older CPUs that are left out by the target change. For that it would make sense to make it optional via a cargo feature.

spfenwick:
That sounds sensible. I'm not exactly thrilled about this change as it is a hack but if it's useful in some situations that's good.

How about "x86-64-sse2-workaround" for the feature name?

I see lint is complaining about the indentation so I'll fix that too and resubmit.

HEnquist:
Sound good. I was thinking of including the rust issue somehow, like "avoid-rustc-issue-116359" or something.

HEnquist:
I had an idea and tried a different approach: https://github.com/HEnquist/camilladsp/pull/303
It's reimplements the biquad calculation using SSE2 intrinsics. The generated code looks ok but I haven't tried it on an Intel cpu yet. I don't see any speed difference vs the generic version.

HEnquist:
@spfenwick it works for me on my ancient debian machine, but could you try it on your system with the crossfeed config that caused trouble also with codegen-units=1?

spfenwick:
I've tested your build with the sse intrinsics and it works well. It fixes the performance issue on the crossfeed config as well as the other configs I tested. It also sounds correct using it in my playback chain.

I compiled with RUSTFLAGS='-C target-cpu=x86-64' cargo build --release

spfenwick:
I've also created a feature to control my "zero sse registers hack" in case you still want it. I'm new to Cargo and Rust so hopefully I've done that correctly.

spfenwick:
On further thought - your approach with the sse2 intrinsics seems like a pretty effective solution. All the issues we've found have pointed to the same Biquad process_chunk function which rustc really hates for some reason. Using the intrinsics avoids the problem pretty cleanly.

HEnquist:
Thanks for testing! I think that intrinsic solution solves it, but there is still a risk that the bug pops up again somewhere else. And if that happens, then the register clearing in this PR might come in handy. I'll merge it and add something in the readme.

PR #319, opened 2024-02-04
Check string pointer after reading clock source name

HEnquist:


HEnquist:
Linux build fails due to https://github.com/dbus2/zbus/pull/568

Issue #382, opened 2024-12-19
Raspberry Pi 4b: glitches when using alsa


yeahnope:
**Describe the bug**
I am using camilladsp in two ways:
 1. usb gadget -> rpi4 -> camilladsp -> usb dac
 2. librespot -> camilladsp -> usb dac

With both configurations I am experiencing glitches (sounds like miss-timed samples) with seemingly random timing (intervals 10-30s) when using alsa directly. This is independent of the chunksize (time between glitches might be longer with larger chunksize, but its hard to tell). Curiously when using pulseaudio before and after camilladsp I don't have this issue. I have tried debugging this issue and found that pulseaudio does two things differently:
 1. By default it uses mmap_interleaved instead of rw_interleaved
 2. As far as I can tell it does not use `snd_pcm_delay()`, but instead `snd_pcm_avail()`.

I have hacked mmap_interleaved support into camilladsp, but found that this did not solve my problems. Using `snd_pcm_avail()` however does fix my issue:
I've just replaced `delay_at_chunk_recvd` with `avail_at_chunk_recvd` and later set  `delay_at_chunk_recvd = (params.chunksize * 4) - avail_at_chunk_recvd` before feeding it to buffer_avg. I know that its not safe to assume that alsa buffer is 4 * the chunksize, so some clean up would be necessary if this were to be made into a PR. 

So my questions would be: Is there a benefit of using delay() over using avail() for syncing input/output? If there's a benefit, would this be something that could be made configurable via the yaml file? 

**To Reproduce**
Use camilladsp with alsa output on raspberry pi 4b

**System info**
Linux 6.6.31-rt30-v8+ #1 SMP PREEMPT_RT Mon Dec  2 13:08:01 Europe 2024 aarch64 GNU/Linux
camilladsp 2.0.3

**Configuration**
```
devices:
  samplerate: 96000
  chunksize: 512
  silence_threshold: -60
  silence_timeout: 3.0
  capture_samplerate: 44100
  adjust_period: 1
  enable_rate_adjust: true
  resampler:
    type: AsyncSinc
    profile: Balanced
  capture:
    type: Alsa
    channels: 2
    device: "hw:Loopback,1"
    format: FLOAT32LE
  playback:
    type: Alsa
    channels: 2
    device: "hw:CARD=D10s,DEV=0"
    format: S32LE

filters:
  gain_for_headroom:
    type: Gain
    parameters:
      gain: -16.0
      scale: dB

  lowpass_fir:
[...]

PR #383, opened 2024-12-28
ALSA: Replace snd_pcm_delay with snd_pcm_avail

yeahnope:
This PR aims at fixing issue #382 

Rpi 4b seems to glitch slightly when repeatedly calling snd_pcm_delay() so we replace it with snd_pcm_avail()


HEnquist:
Thanks for the helpful PR! This was implemented in a slightly different way and included in the 3.0.0 release.

PR #391, opened 2025-01-19
WIP use crossbeam_channel everywhere

HEnquist:


Wang-Yue:
afaik mpsc is implemented using crossbeam?

HEnquist:
Yes the mpsc channel is based on crossbeam_channel, but it doesn't have all the functionality. I needed len() which isn't available in mpsc.

PR #412, opened 2025-04-27
Add MPL 2 license as alternative

HEnquist:
This PR changes the license of CamillaDSP by adding the Mozilla Public License 2.0 as an alternative.

I was recently made aware of the implications of the so called TiVo-clause of the GPL 3 license, which requires sharing any signing keys needed to run modified software on devices distributed with GPL 3 software. This is often a complete showstopper for use in commercial devices, and I don't think this is beneficial for the CamillaDSP project.

GPL also requires all modifications to be shared. This means that for example a Processor or Filter cannot be added without sharing the code, which means that no confidential algorithms can be used. The more permissive MPL 2.0 license on the other hand allows this, a new Processor can be added and used in an embedded device without exposing the algorithm. As long as the new Processor is implemented in a separate file, only the trivial changes to the existing source files need to be shared. This also means that any improvements to the existing functionality (filters, processors etc) would need to be shared.

I think allowing users the option of the MPL 2.0 license is beneficial for CamillaDSP. This license is easier to comply with, and it omits some of the IMO unnecessarily strict terms of GPL. 
Anyone who has contributed, please share your opinion on this topic.
@roderickvd @pavhofman @Wang-Yue @markterrill @benjmarshall @t123yh @balbuze @spfenwick @ElitCoder @JWahle


benjmarshall:
No problem for me, im a big fan of permissive licences.

pavhofman:
Absolutely fine with me, thanks!

balbuze:
Ok for me ! Thanks

HEnquist:
> So, I am heavily in favor of switching to MPL v2 - provided you are O.K. with companies using your software in their proprietary products.

Yes I'm fine with that, it's why I started looking at other licenses. The alternative is either to not allow it at all, which is a bit sad, or to set up some commercial licensing, which I don't really want to deal with.

> 
> You are keeping GPL v3 to make it easier for other GPL projects to include CDSP?

I don't think the GPL is really needed any more, but I don't see any issues with keeping it as an alternative. That way anyone using CamillaDSP today won't need to think about what the new license means for them (unless they want to).

roderickvd:
Sounds great!

Wang-Yue:
LGTM. I love permissive licenses. 
My personal personal projects have always been released in MIT or BSD licenses.

spfenwick:
I'm happy with this change.

HEnquist:
Thanks everyone! I see positive answers from:
@t123yh 
@benjmarshall 
@pavhofman 
@balbuze 
@JWahle 
@roderickvd 
@Wang-Yue 
@spfenwick 

Did I miss anyone? And how about @markterrill and @ElitCoder ?

ElitCoder:
Sounds good to me as well!

HEnquist:
Still waiting for a reply from @markterrill, not sure there will be one since he doesn't seem to have any activity on github since november '23. His changes were some improvements to the websocket interface documentation (and no code changes). I'll wait a few more days and then I'll assume he doesn't mind the added license. In the highly unlikely event that he disagrees, his changes can be removed from the readme.

HEnquist:
I think this is long enough. Merging!

PR #423, opened 2025-12-19
Add native PipeWire audio backend (for next31/v4.0)

Xenograph:
Redo of https://github.com/HEnquist/camilladsp/pull/422 on next31 branch.

I have tested this on my production audio system with no issues. The native PipeWire backend has also reduced latency compared to the ALSA backend for my use case.

# Summary

This PR adds a native PipeWire backend for CamillaDSP on Linux, providing direct integration with PipeWire's graph-based audio routing. This version is rebased on next31 and follows the maintainer's suggestion to use F32 internally (like PulseAudio), simplifying both configuration and code.

  - Adds pipewire-backend Cargo feature
  - Creates filter nodes in the PipeWire graph for capture and playback
  - Supports routing via WirePlumber rules or graphical tools like Helvum
  - Sets NODE_LATENCY property to influence PipeWire quantum scheduling
  - Optional node_name parameter (defaults to camilladsp-capture/camilladsp-playback)

# Changes

  - src/pipewiredevice.rs - New PipeWire capture/playback implementation (~820 lines)
  - src/audiodevice.rs, src/config.rs, src/lib.rs - Integration with existing device infrastructure
  - Cargo.toml - New pipewire-backend feature with pipewire-rs dependencies
  - README.md - Documentation for the new backend and build instructions
  - backend_pipewire.md - Detailed configuration guide with WirePlumber 0.5+ examples

### DIFF
--- a/backend_pipewire.md
+++ b/backend_pipewire.md
@@ -26,8 +26,10 @@ Like PulseAudio, PipeWire uses F32 internally, so no sample format configuration
 capture:
   type: Pipewire
   channels: 2
-  node_name: camilladsp-capture  # optional, this is the default
-  autoconnect_to: null  # optional
+  node_name: camilladsp-capture (*)
+  node_description: CamillaDSP Capture (*)
+  node_group_name: camilladsp (*)
+  autoconnect_to: null (*)
 ```
 
 ### Playback device
@@ -35,9 +37,14 @@ capture:
 playback:
   type: Pipewire
   channels: 2
-  node_name: camilladsp-playback  # optional, this is the default
-  autoconnect_to: null  # optional
+  node_name: camilladsp-playback (*)
+  node_description: CamillaDSP Playback (*)
+  node_group_name: camilladsp (*)
+  autoconnect_to: null (*)
 ```
+
+The parameters marked (*) are optional. The values shown above are the defaults.
+These are used if the parameters are set to `null` or left out from the configuration.
 
 ### Parameters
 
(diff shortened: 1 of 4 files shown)


Xenograph:
Sorry - trusted the LLM a bit too blindly here, figured the logging was a small enough change to not bother testing, but clearly I was wrong.

Pushed a fix. I've tested it and am seeing what appear to be reasonable buffer levels being logged. Have another look when you get a chance :)

HEnquist:
Thanks, I tried it but it still logs only zero. I fixed that here:
https://github.com/Xenograph/camilladsp/pull/1
I also changed some things around to make it more similar to the CoreAudio backend. These two APIs have a lot of similarities and things get easier to maintain if the different backends implement the same functionality in a similar way. There are more areas that can be aligned and I'll continue a bit with that now.

One important bit that is still missing is a copyright message at the top of the pipewire source file. This is needed for release 4.0 that switches to a dual GPLv3 + MPL 2.0 license.

Xenograph:
Also wanted to mention, when running in debug log level, my logs are constantly spammed (like many times per sec) with: `<src/lib.rs:210> The stashed container vector has insufficient capacity, allocating more space`. Is this WAI?

HEnquist:
Vec capacity thing fixed in next31. I meant to push it to my PR #2 also but managed to push it to your PR branch instead.

Xenograph:
Thanks! I've been playing music all day without any hiccups. Just pulled down your latest 2 changes and will continue running with them.

HEnquist:
I have been playing around a bit more as well and things are running well.
I use Helvum to connect the CamillaDSP nodes to others, and then whenever I restart camilla, I have to connect again. That got me thinking that it could be useful to provide an option for letting it connect to some other nodes automatically. The config could have an Option<String> to give he name of a source or sink to connect to. That would make it quite easy to get things running in cases where a very simple graph is enough (which I guess covers quite a lot of users). And leaving it out would make it work like now where it doesn't connect.
Would source/sunk node name be the appropriate property to provide?

HEnquist:
Option 2 looks useful. Connecting to the default devices is probably not that useful.

I suppose node ids are assigned dynamically when nodes are created, so are likely all different after a restart. Node name is better but iiuc pipewire doesn't require node name (and description) to be unique. A bit unfortunate but I don't see anything better than name.

It could make sense to add optional config options for node description and node group, in case anyone would want to run more than one CamillaDSP instance.

HEnquist:
> But the reality is that, today, most apps simply stick to ALSA or Pulse API

Yes exactly. Pretty much all apps will just use their favorite api and output to the default device. Most users want system wide dsp, for stuff like room correction or speaker crossovers. The simplest way to do that is to set some unused or dummy device as the default. And then let CamillaDSP capture from the monitor output of this device, do its thing, and output to the "real" output device. This approach works nicely on Pulse.

HEnquist:
Autoconnect is implemented here:
https://github.com/Xenograph/camilladsp/pull/3
Very little is actually needed. The id in the stream connect() is deprecated, and instead there is a property for the target, called "target.object" (there is a constant for it in pw_sys (PW_KEY_TARGET_OBJECT) but not in pw::keys. I assume it will be added there at some point). And the AUTOCONNECT stream flag is needed.
This way it accepts a node name or serial number (given as a string).

Xenograph:
Looking great! Happy to plan to follow up in a future PR with a tutorial, especially if you want to point to an existing one in particular that we should mirror.

PR #425, opened 2026-01-20
Lookahead Limiter

brainrake:
This PR adds a new filter type: Lookahead Limiter.

It delays the signal by the attack time, detects peaks in advance, uses linear gain reduction ramp 
to reach the required level exactly when the peak arrives. After the peak passes, the gain
is restored using an exponential release curve.

At least one second long buffer is always reserved to avoid runtime buffer growth, attack time is limited to this maximum.

### DIFF
--- a/README.md
+++ b/README.md
@@ -129,6 +129,7 @@ and the Mozilla Public License Version 2.0:
    - **[IIR](#iir)**
    - **[Dither](#dither)**
    - **[Limiter](#limiter)**
+   - **[Lookahead limiter](#lookahead-limiter)**
    - **[Difference equation](#difference-equation)**
 - **[Processors](#processors)**
    - **[Compressor](#compressor)**
@@ -2376,6 +2377,29 @@ Example:
 Parameters:
   * `soft_clip`: enable soft clipping. Set to `false` to use hard clipping. Optional, defaults to `false`.
   * `clip_limit`: the level in dB to clip at.
+
+### Lookahead Limiter
+The "LookaheadLimiter" filter is a lookahead limiter with smooth attack ramping and exponential release.
+It delays the signal by the attack time, detects peaks in advance and smoothly ramps gain reduction
+to reach the required level exactly when the peak arrives. After the peak passes, the gain
+is restored using an exponential release curve.
+
+Example:
+```
+  example_lookahead_limiter:
+    type: LookaheadLimiter
+    parameters:
+      limit: 0.0 (*)
+      attack: 2.0
+      release: 100.0
+```
+
+Parameters:
+  * `limit`: Maximum output level in dB. Optional, defaults to 0.0 dB.
+  * `attack`: Attack/lookahead time in milliseconds. This determines how far ahead the limiter looks for peaks.
+    Must be greater than 0 and less than or equal to 1000 ms.
+  * `release`: Release time in milliseconds. This controls how quickly the gain reduction is released after a peak.
+    Must be greater than or equal to 0 ms.
 
 ### Difference equation
 The "DiffEq" filter implements a generic difference equation filter with transfer function:
--- a/README.md
+++ b/README.md
@@ -2379,10 +2379,7 @@ Parameters:
   * `clip_limit`: the level in dB to clip at.
 
 ### Lookahead Limiter
-The "LookaheadLimiter" filter is a lookahead limiter with smooth attack ramping and exponential release.
-It delays the signal by the attack time, detects peaks in advance and smoothly ramps gain reduction
-to reach the required level exactly when the peak arrives. After the peak passes, the gain
-is restored using an exponential release curve.
+The "LookaheadLimiter" delays the input signal by the attack time, detects peaks in advance and ramps up gain reduction to reach the required level when the peak arrives. After the peak passes, the gain is restored using an exponential release curve.
 
 Example:
 ```
@@ -2390,6 +2387,7 @@ Example:
     type: LookaheadLimiter
     parameters:
       limit: 0.0 (*)
+      unit: ms (*)
       attack: 2.0
       release: 100.0
 ```
--- a/README.md
+++ b/README.md
@@ -2394,8 +2394,9 @@ Example:
 
 Parameters:
   * `limit`: Maximum output level in dB. Optional, defaults to 0.0 dB.
+  * `unit`: Unit for the attack and release times. Required.
   * `attack`: Attack/lookahead time in milliseconds. This determines how far ahead the limiter looks for peaks.
-    Must be greater than 0 and less than or equal to 1000 ms.
+    Must be greater than or eaqual to 0 and less than or equal to 1000 ms.
   * `release`: Release time in milliseconds. This controls how quickly the gain reduction is released after a peak.
     Must be greater than or equal to 0 ms.
 
--- a/README.md
+++ b/README.md
@@ -2396,21 +2396,23 @@ Example:
   example_lookahead_limiter:
     type: LookaheadLimiter
     parameters:
-      limit: 0.0 (*)
+      limit: 0.0
       unit: ms (*)
       attack: 2.0 (*)
-      release: 100.0
+      release: 100.0 (*)
 ```
 
 Parameters:
   * `limit`: Maximum output level in dB. Optional, defaults to 0.0 dB.
-  * `unit`: Unit for the attack and release times. Required.
+  * `unit`: Unit for the attack and release times.
+    Can be `ms`, `us`, `mm` or `samples`.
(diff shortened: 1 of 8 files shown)


HEnquist:
I was thinking a little more about how to handle delays. I think it would be enough if the lookahead limiter calculates the delay in samples from milliseconds the same way as the Delay filter does it. It's almost the same now, just rounding is done differently. That would make it easy to add a matching delay filter to any channel not processed by the limiter.

The readme should of course mention the delay from the limiter, and it would be helpful to include how to set up a delay filter to give the same delay.

HEnquist:
I just remembered that the compressor uses seconds as unit for the attack and release parameters. Since this limiter uses the same parameter names, it would be logical to use the same unit. Then at some point (now or later) they could both be upgraded with a time unit option, like the Delay filter has.

brainrake:
- Rebased on latest next branch.
- Made unit required.
- Fixed algorithm. It's O(n) now and actually works.

brainrake:
I'd like to optimize it a bit more (eg. no heap allocations on each chunk). Also the exponential release config value should behave the same as in the compressor.
Oversampling may be necessary to limit inter-sample peaks (after DAC output).

brainrake:
Fixed: no allocation on chunk anymore (only if chunk size changes).
Fixed: Lookahead Limiter release now behaves the same as Compressor release. 
Complexity is O(lookahead + chunksize) for each chunk. A chunk buffer is allocated on start (and resized on chunk size change). A lookahead delay buffer of 1 sec (the maximum lookahead) is always allocated and filled from input so there are no allocations on parameter change. Only (lookahead + chunksize) samples are read and chunksize samples are written for each chunk.

The gain envelope is generated in two steps.
First, a backwards pass (lookahead + chunksize) generates attack ramps starting (ending) at peaks.
Second, a forward pass (chunksize)  generates exponential release, multiplying by decay coefficient alpha.
The output is computed by multiplying each sample by its gain envelope.

Top. Cant wait to hook it up to FireWire :)

HEnquist:
Alright, I'll take a look!

> and resized on chunk size change

A chunk size change means a complete teardown of the entire pipeline, so you should not expect this to change during the life time of a filter. If you implemented handling of chunk size changes, you can likely simplify things a bit.

HEnquist:
Just a small follow up from my side. I re‑read my earlier comment about a processor style version of this and I realize it was not very clear what I meant.  That question was out of curiosity. I have never used this type of limiter myself, so I was hoping you might have experience with how they are typically implemented. It was not meant as a request to change direction for this PR.  

The current approach is fine for this PR. I was mainly wondering whether a processor version could be a natural next step once this is finished.  And if you need any input from me to move forward, just let me know. No rush. I just wanted to check how things are going on your side.

brainrake:
I have used limiters with lookahead in production and mixing. First time implementing it. Found [this](https://github.com/DanielRudrich/SimpleCompressor/blob/master/docs/lookAheadLimiter.md) useful for overview. Others for details.

Yes, it absolutely makes sense to have a processor, especially for production, mixing, or listening. Also a sidechain could be useful.

For speaker protection in live setups (my current use case), a filter version will do.

I am quite happy with the current version: 

- Time and space complexity for processing a chunk is linear in (chunk + attack) size. It does a backward pass for the attack ramp, a forward pass for the release exponential decay and another forward pass for the gain reduction. Could merge the last two but wanted to keep the code very clear.
- Removed chunk resize logic.
- Unit is required and used for both attack and release. Maybe it should be separate.
- If attack > 1sec, it warns and clamps instead of crashing. Maybe don't need this. But want to have bounded computation per chunk.
- I did some testing, sounded ok, levels looked fine. May need some testing with a scope.

It may be ready for a quick review.

HEnquist:
Thanks, this is in good shape. A few things to fix before merge:

**Delay rounding.** The sample delay needs to round the same way the `Delay` filter does so a
paired delay on the other channels lines up to the sample. `Delay` uses `delay.round() as usize`,
but the limiter does `time_to_samples(...) as usize`, which truncates. Can you switch to `.round()`?
Keeping `time_to_samples` returning the exact float is good, just round it at the call site.

**Release vs compressor.** I checked the math and the release here isn't the same as the
compressor's, even though both are called "release". The compressor uses `exp(-1/n)` (a 1/e time
constant), and this one works out to `exp(-9.21/n)`, since you define release as the time to recover
to roughly -80 dB. That's a fine definition, no need to change the formula. But please document it
in the README so nobody expects the same number to behave identically in both. One sentence saying
the limiter's release is the time to fully recover (to about -80 dB) rather than a time constant is
enough.

**Buffer panic.** `lookahead_buffer.drain(..n)` panics if the chunk is larger than the samplerate
(low rate plus a big chunk). I don't want a config combination that can panic at runtime, so please
bound the buffer with `max(samplerate, chunksize)` (or otherwise guarantee `n` can't exceed its
length).

**attack = 0.** `validate_config` rejects `attack <= 0`, but the README says attack can be 0 and
your peak-limiter test relies on it. So a config with `attack: 0` would fail validation even though
the engine handles it. Change the check to `< 0.0` (or drop the 0 case from the docs and test).

**1 second cap.** The README says attack "must be <= 1000 ms", but the code silently clamps and
warns instead of rejecting. `validate_filter` gets `fs`, so you could validate against the 1 second
bound up front and return a real error. Also the cap is really 1 second, not 1000 ms, once the unit
can be us/mm/samples, so the wording should reflect that.

**Docs.** The attack and release descriptions say "in milliseconds", but `unit` is required and can
be us/ms/mm/samples, so "in the configured unit" is more accurate. Also `unit` is marked `(*)` in
the example but it's required, so the `(*)` should come off (`limit` keeps it, that one defaults to
0.0). And small typo: "eaqual" -> "equal".

**Minor cleanups, optional:**
- `epsilon` is a constant (1e-4), could be a `const` instead of recomputed per instance.
- `configure` calls `time_to_samples(release, ...)` twice, could compute once.

Cheers, Henrik

brainrake:
Thank you for the detailed review! Wow, I missed a lot. Panic, ouch! Updated:
- Attack time rounding works the same as Delay (except it is stored as integer samples since subsample doesn't make a lot of sense here)
- Attack >1sec fails config validation, 0 is ok.
- Release works the same as Compressor (not exactly, Comp works in dB domain and has 1e-6 bias to avoid -Inf; close enough; rabbit hole for later)
- Changed lookahead buffer from VecDeque to LocalRb (like Delay) for faster linear scan.
- Fixed panic by sizing lookahead_buffer to max(1sec, chunksize), added test.
- Updated docs.

brainrake:
Open question: when attack/delay is increased, playback skips back (new_attack - old_attack) samples, which can result in audible pops. The Delay filter instead zeroes the buffer, which can also result in audible pops. Not sure which is preferable. I noticed that Delay clears all samples, which is more than necessary (delay increased from 0.5s to 1.0s -> 1.0s silence, when only 0.5s is strictly needed).

When increasing delay, I reckon zeroing (new_delay - old_delay) samples and keeping the rest may be preferable for both Delay and LookaheadLimiter. This way, no samples are repeated and no samples are thrown away unnecessarily. Not sure how subsamples enter the mix.

According to my research, crossfading old->new buffer and rate limiting the attack change may result in a much less audible parameter change, but this seems somewhat complex and maybe out of scope. In my use case, the param rarely changes, so I don't need this.

HEnquist:
All the fixes look good, verified against the code:
- Attack rounding matches the Delay filter (integer samples, subsample left out, agreed).
- Release now matches the Compressor, with the powf giving a one-pole decay in the dB domain. The 1e-6 / dB caveat in the comment is fine.
- Buffer sized max(1s, chunksize), panic gone, nice regression test.
- attack = 0 accepted, >1s rejected with a real error.

Good tests too, I like the cross-checks against the plain Limiter and the Compressor.

On the open question: clearing the whole buffer on a parameter change is fine. At least for delay filters, the delays are normally small, so the change is audible but not intrusive. Clearing only the (new - old) part is a nice future improvement (and should fix the Delay filter the same way), but not needed now. For crossfading, I don't think the benefit is worth the much increased complexity, so let's skip it.

One thing: could you retarget this to next4.2.0? That release should be ready very soon, so I can merge it straight in.

Cheers,
Henrik

brainrake:
Rebased.
Updated LookaheadLimiter to clear needed part of buffer on reconfigure so it should work similar to Delay.

HEnquist:
Looks good, thanks. Merging into next4.2.0.

brainrake:
Thank you for your attention and guidance, and for creating and maintaining such a gift to the world. I really appreaciate it.

HEnquist:
The new CI failures are on me, they have nothing to do with your changes. Merging!

PR #457, opened 2026-04-09
Ws stream

HEnquist:
Support streaming signal level and state changes over websocket.

### DIFF
--- a/.github/copilot-instructions.md
+++ b/.github/copilot-instructions.md
@@ -19,7 +19,7 @@ This repository is the **CamillaDSP engine** (Rust).
 ## Cargo feature map
 The authoritative feature list is in `Cargo.toml`. When a task depends on optional functionality, check both the feature gate and any platform gate before editing.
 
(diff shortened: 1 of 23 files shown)


Wang-Yue:
I have a proposal: We can add a new API to report device changes. Similar to what has been doing here: https://github.com/Wang-Yue/CamillaDSP-Monitor/blob/main/Sources/CamillaDSPMonitor/Models/AppState%2BDevices.swift#L36-L49

Scenario: Says when the client start the GUI and talk to cdsp, it start to query the list of devices and put them to the selection list. Now with https://github.com/HEnquist/camilladsp/pull/460 they are able to correctly let the user select the sample rate and format. But when user's device list has been changed (such as a new device plugged into the computer), there's no way to update the GUI, unless polling constantly. With the new API to broadcast change, the client program may able to refresh its UI.

HEnquist:
> We can add a new API to report device changes.

Thanks for the proposal! I understand the need to keep a GUI in sync, but my view is that adding device monitoring is out of scope for a dsp engine.
The engine is designed to be lean and focused strictly on audio processing. Incorporating system event monitoring and a broadcast API for device changes would shift the project toward becoming a full-scale backend for graphical applications, which isn't the goal. In my opinion, the responsibility for detecting hardware changes and refreshing the UI should remain with the client-side application.

PR #459, opened 2026-04-10
Threaded alsa shutdown cleanup

siraaris:
## Summary

  Fix ALSA shutdown hangs on `Ctrl-C`.

  This makes shutdown exit cleanly in both:
  - the default ALSA backend
  - the `threaded-alsa` backend

## Context

  This was found while testing CamillaDSP with custom ALSA PCM plugin devices instead of a normal hardware ALSA endpoint.

  The setup included a virtual input path exposed as an ALSA `asym` device and a custom multichannel playback PCM used for AVB output. In that configuration, shutdown behavior was easy to reproduce: a single `Ctrl-C` could leave
  CamillaDSP stuck either in ALSA capture wait/read or in the `threaded-alsa` teardown path.

## What changed

  - make ALSA capture waits interruptible during shutdown
  - add a shared shutdown flag set on exit
  - make processing treat capture-channel disconnect during shutdown as expected
  - add shutdown-aware message forwarding in the threaded ALSA path
  - avoid teardown deadlocks on bounded channels
  - drop playback-inner senders before joining the inner playback thread

  ## Result

  Before:
  - `camilladsp` could hang after `Shutting down`
  - `threaded-alsa` could require a second `Ctrl-C`
  - shutdown could log a spurious processing error

  After:
  - a single `Ctrl-C` exits cleanly
  - no spurious processing error on normal shutdown

  ## Validation

  Tested with:
  ```bash
  cargo build --release
  cargo build --release --features threaded-alsa

  Also verified manually with interactive Ctrl-C on Linux.

### DIFF
--- a/src/alsa_backend/buffermanager.rs
+++ b/src/alsa_backend/buffermanager.rs
@@ -65,7 +65,10 @@ pub trait DeviceBufferManager {
     // Calculate a buffer size and apply it to a hwp container. Only for use when opening a device.
     fn apply_buffer_size(&mut self, hwp: &HwParams) -> Res<()> {
         let min_period = hwp.get_period_size_min().unwrap_or(0);
-        let buffer_frames = self.calculate_buffer_size(min_period);
+        let buffer_frames = self
+            .data()
+            .requested_bufsize
+            .unwrap_or_else(|| self.calculate_buffer_size(min_period));
         let alt_buffer_frames = self.calculate_buffer_size_alt(min_period);
         let data = self.data_mut();
         debug!("Setting buffer size to {buffer_frames} frames");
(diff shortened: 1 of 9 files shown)


HEnquist:
Thanks, this is helpful. It explains the shutdown problem in the threaded ALSA backend nicely.

Several of the challenges also exist in the WASAPI backend, where they are solved differently. I think I’d rather refactor the ALSA backend to behave more like the WASAPI backend, which should simplify shutdown and hopefully make this fix unnecessary. So I’m not planning to merge this as-is, but I do appreciate the work here.

HEnquist:
Can you try this in your setup?
https://github.com/HEnquist/camilladsp/pull/464
I don't have any "difficult" devices on my test system so I'm not quite sure how to test if my version is working or not.

siraaris:
I did a "gh pr checkout 464" and built - Cntr-C seems to shut it down clean and nice.

siraaris:
Superseded by #464

PR #460, opened 2026-04-11
Detailed Hierarchical Hardware Probing

Wang-Yue:
# CamillaDSP Implementation Report: Detailed Hierarchical Hardware Probing

## 1. Problem Description
Previously, CamillaDSP's WebSocket API only returned a list of device names. To determine the supported bit depths for a specific sample rate and channel count, external applications had to perform complex, platform-specific probing. This was inefficient and didn't account for how hardware capabilities often change when switching between Mono, Stereo, or Multi-channel modes.

## 2. Solution: Hierarchical Capability API
We implemented a multi-level probing system that returns a complete hierarchy of hardware capabilities for all major backends (CoreAudio, ALSA, WASAPI, and ASIO).

### 2.1 New WebSocket Commands
- `GetAvailableCaptureDevicesDetailed(backend)`
- `GetAvailablePlaybackDevicesDetailed(backend)`

### 2.2 Hierarchical Data Structure
The API returns a list of devices, where each device provides its capabilities grouped by channel count, then sample rate, then supported bit depths.

```rust
struct AudioDeviceDescriptor {
    name: String,
    description: String,
    capabilities: Vec<ChannelCapability>,
}

struct ChannelCapability {
    channels: usize,
    samplerates: Vec<SamplerateCapability>,
}

struct SamplerateCapability {
    samplerate: usize,
    formats: Vec<String>,
}
```

## 3. Usage Example (JSON Response)
The response explicitly shows the relationship between channel count and supported rates/formats:

```json
{
  "GetAvailablePlaybackDevicesDetailed": {
    "result": "Ok",
    "value": [
      {
        "name": "HDA Intel PCH: ALC285 Analog (hw:0,0)",
        "description": "ALC285 Analog, hw:0,0",
        "capabilities": [
          {
            "channels": 2,
            "samplerates": [
              {
                "samplerate": 44100,
                "formats": ["S16_LE", "S32_LE"]
              },
              {
                "samplerate": 48000,
                "formats": ["S16_LE", "S32_LE", "F32_LE"]
              }
            ]
          },
          {
            "channels": 4,
            "samplerates": [
              {
                "samplerate": 48000,
                "formats": ["S16_LE"]
              }
            ]
          }
        ]
      }
    ]
  }
}
```

## 4. Platform-Specific Probing Logic
- **ALSA**: Probes common channel counts (1, 2, 4, 8, etc.) and refines hardware parameters for each to find valid rate/format combinations.
- **CoreAudio**: Scans physical stream descriptions and groups them by `mChannelsPerFrame`. Includes robust fallbacks for fixed-format devices like microphones.
- **WASAPI**: Probes common channel counts and rates in **Shared Mode** to find compatible configurations with the Windows Audio Engine.
- **ASIO**: Queries driver-level input/output channel counts and associates them with supported sample rates.

## 5. Key Benefits
1.  **True Hardware Insights**: Captures the fact that some devices only support high sample rates or specific bit depths in certain channel configurations (e.g., Stereo vs 7.1).
2.  **Zero Client-Side Logic**: Frontend apps (like the Monitor) can now build their entire "Device Setup" UI by simply traversing a single JSON tree.
3.  **Cross-Platform Consistency**: A unified schema for Linux, macOS, and Windows.

---
**Files Modified:** `src/lib.rs`, `src/socketserver.rs`, `src/coreaudio_backend/device.rs`, `src/alsa_backend/utils.rs`, `src/wasapi_backend/device.rs`, `src/asio_backend/device.rs`.

### DIFF
--- a/src/coreaudio_backend/device.rs
+++ b/src/coreaudio_backend/device.rs
@@ -189,202 +189,199 @@ pub fn list_available_devices(input: bool) -> Vec<(String, String)> {
     names.iter().map(|n| (n.clone(), n.clone())).collect()
 }
 
-pub fn list_available_devices_detailed(input: bool) -> Vec<crate::AudioDeviceDescriptor> {
-    let mut descriptors = Vec::new();
-    if let Ok(all_ids) = get_audio_device_ids() {
-        for device_id in all_ids.iter() {
-            if let Ok(name) = get_device_name(*device_id)
-                && device_supports_scope(*device_id, input)
-            {
-                // Map: Channels -> SampleRate -> Formats
-                let mut capabilities_map: std::collections::HashMap<
-                    usize,
-                    std::collections::HashMap<usize, Vec<String>>,
-                > = std::collections::HashMap::new();
+pub fn get_device_capabilities(
+    device_name: &str,
+    input: bool,
+) -> Result<crate::AudioDeviceDescriptor, crate::DeviceError> {
+    let device_id = match get_device_id_from_name_and_scope(device_name, input) {
+        Some(id) => id,
+        None => {
+            return Err(crate::DeviceError::DeviceNotFound(device_name.to_string()));
+        }
+    };
 
-                if let Ok(supported_formats) = get_supported_physical_stream_formats(*device_id) {
-                    for ranged_desc in supported_formats {
-                        let asbd = ranged_desc.mFormat;
-                        let channels = asbd.mChannelsPerFrame as usize;
+    if !device_supports_scope(device_id, input) {
+        return Err(crate::DeviceError::DeviceNotFound(device_name.to_string()));
+    }
 
-                        // Get format string
-                        let is_float = (asbd.mFormatFlags & kAudioFormatFlagIsFloat) != 0;
-                        let bit_depth = asbd.mBitsPerChannel;
-                        let format_str = if is_float {
-                            if bit_depth == 64 { "F64" } else { "F32" }
-                        } else {
-                            match bit_depth {
-                                16 => "S16",
-                                24 => "S24",
-                                32 => "S32",
-                                _ => continue,
-                            }
-                        };
+    let mut capabilities_map: std::collections::HashMap<
+        usize,
+        std::collections::HashMap<usize, Vec<String>>,
+    > = std::collections::HashMap::new();
 
-                        let rates = if ranged_desc.mSampleRateRange.mMinimum
-                            == ranged_desc.mSampleRateRange.mMaximum
-                        {
-                            vec![ranged_desc.mSampleRateRange.mMinimum as usize]
-                        } else {
-                            vec![
-                                ranged_desc.mSampleRateRange.mMinimum as usize,
-                                ranged_desc.mSampleRateRange.mMaximum as usize,
-                            ]
-                        };
+    if let Ok(supported_formats) = get_supported_physical_stream_formats(device_id) {
+        for ranged_desc in supported_formats {
+            let asbd = ranged_desc.mFormat;
+            let channels = asbd.mChannelsPerFrame as usize;
 
-                        let rate_map = capabilities_map.entry(channels).or_default();
-                        for rate in rates {
-                            rate_map
-                                .entry(rate)
-                                .or_default()
-                                .push(format_str.to_string());
-                        }
-                    }
+            // Get format string
+            let is_float = (asbd.mFormatFlags & kAudioFormatFlagIsFloat) != 0;
+            let bit_depth = asbd.mBitsPerChannel;
+            let format_str = if is_float {
+                if bit_depth == 64 { "F64" } else { "F32" }
+            } else {
+                match bit_depth {
+                    16 => "S16",
+                    24 => "S24",
+                    32 => "S32",
+                    _ => continue,
                 }
+            };
 
-                if capabilities_map.is_empty() {
-                    // Fallback to nominal rates and current config
-                    let property_address = AudioObjectPropertyAddress {
-                        mSelector:
-                            objc2_core_audio::kAudioDevicePropertyAvailableNominalSampleRates,
-                        mScope: kAudioObjectPropertyScopeGlobal,
-                        mElement: kAudioObjectPropertyElementMain,
-                    };
-                    let mut data_size = 0u32;
-                    if unsafe {
-                        AudioObjectGetPropertyDataSize(
-                            *device_id,
-                            NonNull::from(&property_address),
-                            0,
-                            null(),
-                            NonNull::from(&mut data_size),
-                        )
-                    } == 0
-                    {
-                        let nbr_items = data_size
-                            / mem::size_of::<objc2_core_audio_types::AudioValueRange>() as u32;
-                        let mut ranges =
-                            vec![
-                                unsafe { mem::zeroed::<objc2_core_audio_types::AudioValueRange>() };
-                                nbr_items as usize
-                            ];
-                        if unsafe {
-                            AudioObjectGetPropertyData(
-                                *device_id,
-                                NonNull::from(&property_address),
-                                0,
-                                null(),
-                                NonNull::from(&mut data_size),
-                                NonNull::new(ranges.as_mut_ptr()).unwrap().cast(),
-                            )
-                        } == 0
-                        {
-                            let current_format = get_current_device_format(*device_id, input)
-                                .unwrap_or_else(|_| "F32".to_string());
+            let rates =
+                if ranged_desc.mSampleRateRange.mMinimum == ranged_desc.mSampleRateRange.mMaximum {
+                    vec![ranged_desc.mSampleRateRange.mMinimum as usize]
+                } else {
+                    vec![
+                        ranged_desc.mSampleRateRange.mMinimum as usize,
+                        ranged_desc.mSampleRateRange.mMaximum as usize,
+                    ]
+                };
 
-                            // Probe channel count
-                            let scope = if input {
-                                kAudioObjectPropertyScopeInput
-                            } else {
-                                kAudioObjectPropertyScopeOutput
-                            };
-                            let config_address = AudioObjectPropertyAddress {
-                                mSelector: kAudioDevicePropertyStreamConfiguration,
-                                mScope: scope,
-                                mElement: kAudioObjectPropertyElementWildcard,
-                            };
-                            let mut config_size = 0u32;
-                            let mut channels = 2; // fallback
-                            if unsafe {
-                                AudioObjectGetPropertyDataSize(
-                                    *device_id,
-                                    NonNull::from(&config_address),
-                                    0,
-                                    null(),
-                                    NonNull::from(&mut config_size),
-                                )
-                            } == 0
-                            {
-                                // Allocate aligned storage for AudioBufferList
-                                let layout = std::alloc::Layout::from_size_align(
-                                    config_size as usize,
-                                    mem::align_of::<AudioBufferList>(),
-                                )
-                                .unwrap();
-                                let buf_list_ptr = unsafe { std::alloc::alloc_zeroed(layout) }
-                                    as *mut AudioBufferList;
-                                if !buf_list_ptr.is_null() {
-                                    if unsafe {
-                                        AudioObjectGetPropertyData(
-                                            *device_id,
-                                            NonNull::from(&config_address),
-                                            0,
-                                            null(),
-                                            NonNull::from(&mut config_size),
-                                            NonNull::new(buf_list_ptr).unwrap().cast(),
-                                        )
-                                    } == 0
-                                    {
-                                        let nbr_buffers =
-                                            unsafe { (*buf_list_ptr).mNumberBuffers } as usize;
-                                        // mBuffers is a variable-length array starting at the first element
-                                        let buffers_ptr =
-                                            unsafe { (*buf_list_ptr).mBuffers.as_ptr() };
-                                        let buffers = unsafe {
-                                            std::slice::from_raw_parts(buffers_ptr, nbr_buffers)
-                                        };
-                                        channels =
-                                            buffers.iter().map(|b| b.mNumberChannels).sum::<u32>()
-                                                as usize;
-                                    }
-                                    unsafe { std::alloc::dealloc(buf_list_ptr as *mut u8, layout) };
-                                }
-                            }
-
-                            let rate_map = capabilities_map.entry(channels).or_default();
-                            for range in ranges {
-                                rate_map
-                                    .entry(range.mMinimum as usize)
-                                    .or_default()
-                                    .push(current_format.clone());
-                            }
-                        }
-                    }
-                }
-
-                let mut capabilities: Vec<crate::ChannelCapability> = capabilities_map
-                    .into_iter()
-                    .map(|(channels, rate_map)| {
-                        let mut samplerates: Vec<crate::SamplerateCapability> = rate_map
-                            .into_iter()
-                            .map(|(samplerate, mut formats)| {
-                                formats.sort_unstable();
-                                formats.dedup();
-                                crate::SamplerateCapability {
-                                    samplerate,
-                                    formats,
-                                }
-                            })
-                            .collect();
-                        samplerates.sort_unstable_by_key(|s| s.samplerate);
-                        crate::ChannelCapability {
-                            channels,
-                            samplerates,
-                        }
-                    })
-                    .collect();
-
-                capabilities.sort_unstable_by_key(|c| c.channels);
-
-                descriptors.push(crate::AudioDeviceDescriptor {
-                    name: name.to_string(),
-                    description: name.to_string(),
-                    capabilities,
-                });
+            let rate_map = capabilities_map.entry(channels).or_default();
+            for rate in rates {
+                rate_map
+                    .entry(rate)
+                    .or_default()
+                    .push(format_str.to_string());
             }
         }
     }
-    descriptors
+
+    if capabilities_map.is_empty() {
+        // Fallback to nominal rates and current config
+        let property_address = AudioObjectPropertyAddress {
+            mSelector: objc2_core_audio::kAudioDevicePropertyAvailableNominalSampleRates,
+            mScope: kAudioObjectPropertyScopeGlobal,
+            mElement: kAudioObjectPropertyElementMain,
+        };
+        let mut data_size = 0u32;
+        if unsafe {
+            AudioObjectGetPropertyDataSize(
+                device_id,
+                NonNull::from(&property_address),
+                0,
+                null(),
+                NonNull::from(&mut data_size),
+            )
+        } == 0
+        {
+            let nbr_items =
+                data_size / mem::size_of::<objc2_core_audio_types::AudioValueRange>() as u32;
+            let mut ranges =
+                vec![
+                    unsafe { mem::zeroed::<objc2_core_audio_types::AudioValueRange>() };
+                    nbr_items as usize
+                ];
+            if unsafe {
+                AudioObjectGetPropertyData(
+                    device_id,
+                    NonNull::from(&property_address),
+                    0,
+                    null(),
+                    NonNull::from(&mut data_size),
+                    NonNull::new(ranges.as_mut_ptr()).unwrap().cast(),
+                )
+            } == 0
+            {
+                let current_format = get_current_device_format(device_id, input)
+                    .unwrap_or_else(|_| "F32".to_string());
+
+                // Probe channel count
+                let scope = if input {
+                    kAudioObjectPropertyScopeInput
+                } else {
+                    kAudioObjectPropertyScopeOutput
+                };
+                let config_address = AudioObjectPropertyAddress {
+                    mSelector: kAudioDevicePropertyStreamConfiguration,
+                    mScope: scope,
+                    mElement: kAudioObjectPropertyElementWildcard,
+                };
+                let mut config_size = 0u32;
+                let mut channels = 2; // fallback
+                if unsafe {
+                    AudioObjectGetPropertyDataSize(
+                        device_id,
+                        NonNull::from(&config_address),
+                        0,
+                        null(),
+                        NonNull::from(&mut config_size),
+                    )
+                } == 0
+                {
+                    // Allocate aligned storage for AudioBufferList
+                    let layout = std::alloc::Layout::from_size_align(
+                        config_size as usize,
+                        mem::align_of::<AudioBufferList>(),
+                    )
+                    .unwrap();
+                    let buf_list_ptr =
+                        unsafe { std::alloc::alloc_zeroed(layout) } as *mut AudioBufferList;
+                    if !buf_list_ptr.is_null() {
+                        if unsafe {
+                            AudioObjectGetPropertyData(
+                                device_id,
+                                NonNull::from(&config_address),
+                                0,
+                                null(),
+                                NonNull::from(&mut config_size),
+                                NonNull::new(buf_list_ptr).unwrap().cast(),
+                            )
+                        } == 0
+                        {
+                            let nbr_buffers = unsafe { (*buf_list_ptr).mNumberBuffers } as usize;
+                            // mBuffers is a variable-length array starting at the first element
+                            let buffers_ptr = unsafe { (*buf_list_ptr).mBuffers.as_ptr() };
+                            let buffers =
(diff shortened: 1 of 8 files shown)


Wang-Yue:
Background: without this api client program will have to make a lot of effort to reinvent the wheel, see https://github.com/Wang-Yue/CamillaDSP-Monitor/commit/80d2628da1b007ce592233dfa6618281584cddf4

tested: CoreAudio, ALSA, WASAPI
not tested: ASIO

HEnquist:
Please don't make very big PRs like this without any prior discussion of the design. I know it's tempting to let an LLM make very big changes since it does them so quickly, but please open an issue first so we can hash out the general design before doing any implementation.
I don't mind adding this functionality, but there are some quite serious issues here.
First of all, it's really not good to have a single call to fetch details for all devices. There may be many, and for example with Wasapi it is quite expensive to probe device capabilities. The command should take a device name, and only report probe that particular device. On Wasapi, ASIO and Alsa you need to open the device to query, which means any device that is busy cannot be probed. On ASIO there is the additional complication that an app can only access one device at a time. So if CamillaDSP is running with ASIO, probing is unavailable. So there have to be a few different error responses. At least for device not found, device busy, and some general error.
Finally I see the response includes raw strings for sample format. It should use the same formats as in the config. And the list of common sample formats is repeated many times over.

Wang-Yue:
Reviewer comments fixed.

HEnquist:
The updates partly addressed my initial set of comments. I have looked more closely now, and found these things that needs to be sorted out.

1. WASAPI capability probing is currently shared-mode-only, and the reply format does not indicate which mode a returned configuration belongs to. The new probe in `src/wasapi_backend/device.rs:111-220` hard-codes `ShareMode::Shared` for every test at `src/wasapi_backend/device.rs:193-199`. That already means the result says nothing about exclusive mode, even though the actual runtime path explicitly switches between shared and exclusive in `src/wasapi_backend/device.rs:369-410`. I think the capability reply needs to indicate whether a given returned configuration is for shared or exclusive mode. Otherwise a client cannot tell whether a reported format/rate/channel combination is actually usable for the mode it intends to open.

2. The current WASAPI shared-mode probe can also misreport integer formats as supported when it only tested float. In the shared branch of `src/wasapi_backend/device.rs:317-335`, `get_supported_wave_format` ignores the requested `WasapiSampleFormat` and always probes `F32_LE`. But the outer loop in `src/wasapi_backend/device.rs:183-203` still iterates over `S16`, `S24`, `S32`, and `F32`, and records the corresponding format name on success. In practice this means the same successful shared-mode float probe can be labeled as support for integer formats that were never actually tested. Suggested fix: either model shared mode explicitly as float-only and only report that, or make the capability model include mode and reuse the same share-mode-aware logic as the existing open path.

3. CoreAudio capability probing collapses sample-rate ranges into incomplete discrete results. In `src/coreaudio_backend/device.rs:192-225`, the physical-format path keeps only the minimum and maximum of a reported range at `src/coreaudio_backend/device.rs:209-217`. In the fallback path in `src/coreaudio_backend/device.rs:227-280`, it only keeps the minimum at `src/coreaudio_backend/device.rs:273-277`. That means a device advertising a continuous or broad nominal range can be reported as supporting only a tiny subset of what it actually supports. Since the websocket/API schema currently returns discrete sample rates, I think the implementation needs to either expand ranges against a shared curated rate table or change the schema to represent ranges directly.

4. The ASIO capability probe resolves sample format from the input side even for playback requests. In `src/asio_backend/device.rs:1219-1288`, `resolve_format(&None, true)` is always used at `src/asio_backend/device.rs:1252-1255`. For playback requests this should be direction-aware. On devices where input and output formats differ, `GetPlaybackDeviceCapabilities` can report the wrong format. Suggested fix: pass the requested direction through to `resolve_format`, and return an error if format detection fails rather than continuing with partial data.

5. The ASIO probe also silently suppresses format-detection failures and can serialize incomplete capability entries. In `src/asio_backend/device.rs:1252-1255`, failure to resolve the format is ignored, and the code continues building capabilities with an empty `formats` list in `src/asio_backend/device.rs:1269-1274`. I think this should return a typed `DeviceError` instead. A capability reply with sample rates but no formats is hard for a client to interpret and likely not actionable.

6. The new typed device-error mapping is based on string matching on formatted backend errors rather than real backend error codes or variants. This is new in ALSA in `src/alsa_backend/utils.rs:175-185`, WASAPI in `src/wasapi_backend/device.rs:160-168`, and ASIO in `src/asio_backend/device.rs:1235-1242`. Since the websocket API now exposes distinct error classes like not found vs busy, this becomes part of the public contract. Matching on formatted strings is brittle across wrapper versions and localization. I think it would be safer either to inspect actual backend error codes/variants or to fall back to a generic device error when a stable classification is not available.

7. This PR adds more duplication around capability policy, especially sample-rate tables. There is already an ASIO common-rates list in the existing open path at `src/asio_backend/device.rs:1052-1060`, and this PR adds another one in `src/asio_backend/device.rs:1223-1226`, plus another in `src/wasapi_backend/device.rs:179-182`, while ALSA already has `STANDARD_RATES` in `src/alsa_backend/utils.rs:35-39`. I would strongly prefer one shared source for the curated rate list and, where possible, shared helpers for shaping capability results so the backends do not drift.

8. The new CoreAudio probing code is also fairly large and seems to duplicate logic instead of reusing existing CoreAudio helper paths. The new helper set starts at `src/coreaudio_backend/device.rs:192`, while the existing device-open path already uses the binding helpers in `src/coreaudio_backend/device.rs:499-621`. I would suggest refactoring this so capability probing reuses the same lower-level logic already trusted for actual device setup, rather than adding a second independent interpretation path.

Wang-Yue:
Thanks for the feedback. Unfortunately I cannot access any of your links --- links such as "https://github.com/HEnquist/camilladsp/pull/src/wasapi_backend/device.rs#L104" always give me blank page. I also manually check the file at the given line, and it doesn't matches what you said in any case. Can you use link format that can be accessed by anyone, such as "https://github.com/Wang-Yue/camilladsp/blob/1037d53cdd92abf9ae844e10bfcb29e2972ff086/src/coreaudio_backend/device.rs#L79" so I can get what you mean?

HEnquist:
Sorry the URLs were all wrong. I edited to just include the relative path of the source files and (corrected) line number.

Wang-Yue:
Thanks for your detailed review and fixing the line numbers. I addressed the issues you raised. PTAL

src/coreaudio_backend/device.rs
```diff
@@ -186,35 +188,274 @@ pub fn list_available_devices(input: bool) -> Vec<(String, String)> {
...
+
+            let rate_map = capabilities_map.entry(channels).or_default();
+            for rate in rates {
+                rate_map.entry(rate).or_default().push(format_str.clone());
+            }
+        }
+    }
+}
+
+fn get_fallback_capabilities(
```
HEnquist: What concrete CoreAudio failure mode is this fallback meant to handle? Do you know of any real devices or drivers where the physical-format query in src/coreaudio_backend/device.rs:191 returns nothing, but the device is still usable?

HEnquist:
I looked closer at the CoreAudio part only together with copilot:

Findings

High: The new fallback path can advertise capabilities that the open path still cannot actually honor. [device.rs:226](vscode-file://vscode-app/Applications/Visual%20Studio%20Code.app/Contents/Resources/app/out/vs/code/electron-browser/workbench/workbench.html) synthesizes capabilities from nominal sample-rate ranges plus the current format, but opening with an explicit format still requires physical-format matching in [device.rs:461](vscode-file://vscode-app/Applications/Visual%20Studio%20Code.app/Contents/Resources/app/out/vs/code/electron-browser/workbench/workbench.html) and [device.rs:552](vscode-file://vscode-app/Applications/Visual%20Studio%20Code.app/Contents/Resources/app/out/vs/code/electron-browser/workbench/workbench.html). That means the PR can now report format/rate combinations as “capabilities” even when the backend still has no way to validate or set them via physical formats. If there is no concrete CoreAudio device motivating this, I would remove the fallback entirely.

High: [device_supports_scope](vscode-file://vscode-app/Applications/Visual%20Studio%20Code.app/Contents/Resources/app/out/vs/code/electron-browser/workbench/workbench.html) is now materially less trustworthy than on master. On master, it only checked whether the stream-configuration property existed and had at least one buffer, and returned false on query failure. In this PR it delegates to [device.rs:290](vscode-file://vscode-app/Applications/Visual%20Studio%20Code.app/Contents/Resources/app/out/vs/code/electron-browser/workbench/workbench.html), which defaults to 2 channels on failure and then treats > 0 as supported in [device.rs:427](vscode-file://vscode-app/Applications/Visual%20Studio%20Code.app/Contents/Resources/app/out/vs/code/electron-browser/workbench/workbench.html). That can turn CoreAudio query failures into false positives for scope support, which affects both device listing and capability lookup. If the fallback path goes away, I would restore the old simpler [device_supports_scope](vscode-file://vscode-app/Applications/Visual%20Studio%20Code.app/Contents/Resources/app/out/vs/code/electron-browser/workbench/workbench.html) logic.

Medium: Most of the added low-level CoreAudio plumbing exists only to support the speculative fallback, not the actual new feature. Compared with master, the genuinely required addition is the new CoreAudio [get_device_capabilities](vscode-file://vscode-app/Applications/Visual%20Studio%20Code.app/Contents/Resources/app/out/vs/code/electron-browser/workbench/workbench.html) entry point in [device.rs:342](vscode-file://vscode-app/Applications/Visual%20Studio%20Code.app/Contents/Resources/app/out/vs/code/electron-browser/workbench/workbench.html), because it is now wired through [lib.rs:517](vscode-file://vscode-app/Applications/Visual%20Studio%20Code.app/Contents/Resources/app/out/vs/code/electron-browser/workbench/workbench.html) and used by the websocket layer. But [get_current_device_format](vscode-file://vscode-app/Applications/Visual%20Studio%20Code.app/Contents/Resources/app/out/vs/code/electron-browser/workbench/workbench.html) at [device.rs:389](vscode-file://vscode-app/Applications/Visual%20Studio%20Code.app/Contents/Resources/app/out/vs/code/electron-browser/workbench/workbench.html), [get_device_channel_count](vscode-file://vscode-app/Applications/Visual%20Studio%20Code.app/Contents/Resources/app/out/vs/code/electron-browser/workbench/workbench.html) at [device.rs:290](vscode-file://vscode-app/Applications/Visual%20Studio%20Code.app/Contents/Resources/app/out/vs/code/electron-browser/workbench/workbench.html), and most of the new imports are only needed because of the fallback. If the goal is to keep this change minimal, those look removable.

Medium: The current structure makes the file look more “careful” than it really is. [get_physical_capabilities](vscode-file://vscode-app/Applications/Visual%20Studio%20Code.app/Contents/Resources/app/out/vs/code/electron-browser/workbench/workbench.html) in [device.rs:191](vscode-file://vscode-app/Applications/Visual%20Studio%20Code.app/Contents/Resources/app/out/vs/code/electron-browser/workbench/workbench.html) is a reasonable extraction of the real capability logic, and [capabilities_from_map](vscode-file://vscode-app/Applications/Visual%20Studio%20Code.app/Contents/Resources/app/out/vs/code/electron-browser/workbench/workbench.html) is fine. The complexity spike comes from trying to recover from missing physical-format data rather than from the capability feature itself. I would keep the physical-path extraction and map conversion, but question the value of the fallback machinery unless the author can name a real device class that needs it.

Open Questions

Do we have any actual CoreAudio devices or drivers where the physical-format query fails or returns nothing, but the device is still usable enough that this fallback improves behavior rather than fabricating it?
If not, why not keep the implementation limited to the physical-format path and return empty capabilities when CoreAudio does not expose them?
Summary

Compared with master, not all of the added code looks necessary. The new CoreAudio capability API itself appears needed because the PR adds cross-backend capability reporting, but the fallback path looks like the part most likely added without a concrete requirement. If you want to push back on over-complexity, I think the strongest position is: keep the new capability entry point and the extracted physical-format parsing, remove the fallback path, and restore the old simple scope check.

src/coreaudio_backend/device.rs
Wang-Yue: Done.

HEnquist: This was a question, I assumed you had some reason for adding the fallback but I could not see what it was.

Wang-Yue: Your comment is very valid. I do have a pair of weird bluetooth earbuds that erroneously reports back an empty list and the fallback code can stuff a capability setting. But even with that setting I cannot make it work.

Wang-Yue:
To answer your question --- there are devices that camilladsp outputs empty capability with the current infra, but can get extra supported capabilities with get_fallback_capabilities. But camilladsp is not able to open the device with a supported capability. It seems camilladsp (and under the hood, coreaudio-rs) has difficulty supporting certain bluetooth devices.  I will file an issue shortly. 

I highly recommend accepting this PR as is as it's already outputting the correct json of the device capabilities camilladsp supports right now, and my git branch is already full of changes and merge conflicts. We can revisit this again when coreaudio-rs is updated with the right fix. And if you take a look at the proposed diff in that issue, you'll find the changes to the camilladsp code is very small. Most of the changes will be in your coreaudio-rs.

src/coreaudio_backend/device.rs
Wang-Yue: See the latest comment. It turned out camilladsp's current code doesn't support my bluetooth device playback in the first place. So even if get_fallback_capabilities gets back the correct data, camilladsp is still not able to use it as the device setting. I recommend having this PR accepted, and we can fix the underlying issue in another PR.

HEnquist:
> It seems camilladsp (and under the hood, coreaudio-rs) has difficulty supporting certain bluetooth devices. I will file an issue shortly.

Thanks, I saw the new issue and that explains it perfectly.



> Most of the changes will be in your coreaudio-rs.

Will you submit your changes to the coreaudio-rs repo? I'm sure the maintainer will appreciate it and he is is usually quick to respond.



> I highly recommend accepting this PR as is as it's already outputting the correct json of the device capabilities camilladsp supports right now

Sorry but I don't see this as ready for merging yet. There are some things is the Wasapi code I have some doubts about and would like to take a proper look at. And the ASIO changes are still completely untested, which is problematic since the ASIO api is quite finicky.

Wang-Yue:
> Will you submit your changes to the coreaudio-rs repo? I'm sure the maintainer will appreciate it and he is is usually quick to respond.

I submitted it. StreamID/DeviceID is not the only issue --- the APIs do not take into account that certain devices support different capabilities in its input and output. Since this is api breaking change I don't know how far the change can go. 

> Sorry but I don't see this as ready for merging yet. There are some things is the Wasapi code I have some doubts about and would like to take a proper look at. And the ASIO changes are still completely untested, which is problematic since the ASIO api is quite finicky.

macOS/CoreAudio is the only platform I have a lot of experience with. I have zero knowledge on Rust. To work on this I had to set up dev enviroment for 3 platforms and two virtual machines on my Mac, plus multiple git branches (one is your ws_stream, which has a lot of merge conflict with this), and it gets messy pretty fast. 

It's my bad that I shouldn't take something more than what my capability allows in the first place. Maybe next time I should only do CoreAudio changes first and leave other platforms for later. I would recommend following your previous way to set up a nextXX branch so we have a single source of truth, and we can check in code regularly, and fix remaining issues in that branch later.

HEnquist:
> I submitted it. StreamID/DeviceID is not the only issue --- the APIs do not take into account that certain devices support different capabilities in its input and output. Since this is api breaking change I don't know how far the change can go.

Great!



> macOS/CoreAudio is the only platform I have a lot of experience with. I have zero knowledge on Rust. To work on this I had to set up dev enviroment for 3 platforms and two virtual machines on my Mac, plus multiple git branches (one is your ws_stream, which has a lot of merge conflict with this), and it gets messy pretty fast.

Yeah working on changes that touch several platforms is quite time consuming. I started with virtual machines but that got too annoying, so now I simply use three different computers.

> 
> It's my bad that I shouldn't take something more than what my capability allows in the first place. Maybe next time I should only do CoreAudio changes first and leave other platforms for later. I would recommend following your previous way to set up a nextXX branch so we have a single source of truth, and we can check in code regularly, and fix remaining issues in that branch later.

I have created the `next4.2.0` branch and moved the PR to that one. The changes in the `ws_stream` branch are in many of the same files as this one, but I would expect them to be quite easy to resolve. An LLM should do it quickly. I don't plan any other major changes at the moment, so there should no other conflicts than the ones with `ws_stream`.

I hope to get a chance to look at wasapi and Asio very soon.

HEnquist:
Findings for Wasapi:

## Findings

1. High: the exclusive-mode probe is bounded by the shared-mode mix format channel count, which is not a valid upper bound for exclusive support. In `src/wasapi_backend/device.rs#L167`, the code uses `get_mixformat().get_nchannels()` and then probes `1..=max_channels`. `GetMixFormat` describes the audio engine's shared-mode mix format, not the device's maximum exclusive-mode channel capability. This can underreport valid exclusive configurations.

2. Medium: the code does preserve some shared-versus-exclusive information, but only through a string-encoding hack in the reported sample format values. That is better than a fully flattened matrix, but it is still the wrong abstraction. It makes clients parse semantics out of strings, is backend-specific, and keeps the API lossy and brittle. The problem is still in the response shape around the capability construction in `src/wasapi_backend/device.rs#L181` through `src/wasapi_backend/device.rs#L205`: mode is being smuggled through `formats` instead of being represented structurally.

3. Medium: shared-mode capability detection should not be modeled as a grid probe over channels, rates, and `F32`. In `src/wasapi_backend/device.rs#L172` through `src/wasapi_backend/device.rs#L191`, the code probes shared mode by trying `F32` over the standard rate table and the probed channel range. For WASAPI, the authoritative shared-mode format is the mix format. Using the mix format directly is a more accurate model for the shared path CamillaDSP actually uses, and avoids reporting a synthetic shared-mode matrix.

## Suggested response shape

A shape like this looks reasonable:

- `AudioDeviceDescriptor`
- `name`
- `description`
- `capability_sets: Vec<DeviceCapabilitySet>`

Where each `DeviceCapabilitySet` contains:

- `mode`
- `capabilities: Vec<ChannelCapability>`

And `mode` is one of:

- `Shared`
- `Exclusive`
- `Unified`

This works well across backends:

- WASAPI can return one `Shared` set based on `GetMixFormat`, plus one `Exclusive` set based on explicit probing.
- ALSA, CoreAudio, and ASIO can return a single `Unified` set using the existing capability matrix.
- Clients get one consistent schema and can render either one section or several without backend-specific parsing hacks.

## Summary

The current WASAPI implementation is on the right track, but the response model is still doing too much implicit encoding. The main technical issue is that exclusive probing is limited by a shared-mode property. The main API issue is that shared-versus-exclusive is not represented explicitly, and is instead partially encoded in format strings. A `DeviceCapabilitySet` layer fixes that cleanly and also generalizes well to the other backends.

HEnquist:
And for ASIO:

- High: src/asio_backend/device.rs:1215-1284, src/asio_backend/device.rs:817-820, src/asio_backend/device.rs:803-811, src/asio_backend/device.rs:136-138, src/asio_backend/device.rs:192

  The ASIO capability probe is not side-effect free. `get_device_capabilities()` calls `load_driver_by_name()`, and `load_driver_by_name()` unconditionally calls `teardown_asio_driver()` first. That means a capability query can tear down the currently loaded ASIO driver even when CamillaDSP is already running playback or capture. Since the backend already tracks in-process ASIO ownership, this path should first detect that state and return `DeviceBusy` instead of touching the loaded driver.

- High: src/asio_backend/device.rs:1257-1284

  `ASIOGetChannels()` failure is silently converted into a successful empty descriptor. If `ASIOGetChannels()` returns a non-zero error code, the code just skips capability construction, tears the driver down, and still returns `Ok(AudioDeviceDescriptor { capabilities: [] })`. That hides real probe failures and will misreport broken or unavailable drivers as "device exists but has no capabilities". This should return an error instead.

- Medium: src/asio_backend/device.rs:1258-1275

  The function can publish a `ChannelCapability` with `channels = 0`. If a driver only supports the opposite direction, `supported_rates` can still be non-empty and this code will emit a zero-channel capability entry. That is not a meaningful capability and should be filtered out or treated as unsupported for the requested direction.

- Medium: src/asio_backend/device.rs:1227-1229, src/asio_backend/device.rs:836-846

  The current error mapping is too coarse to support "busy because another process is using the device". `load_asio_driver()` only gives a boolean and `ASIOInit()` gives a backend-specific integer, but this code does not decode those into a reliable external-busy signal. `DeviceNotFound` is already handled from the enumerated driver names, but the remaining load/init failures are not detailed enough to safely map to `DeviceBusy`. Those should stay as `Other` unless you add a validated busy mapping.

- Medium: src/asio_backend/device.rs:1219-1229

  Because the probe immediately loads the driver once the name matches, it currently misses the chance to return `DeviceBusy` for the in-process case before disturbing shared ASIO state. Even if you keep the generic `Other` mapping for external failures, the in-process busy case should still be handled explicitly before calling into the loader.

HEnquist:
You are using debug formatting to serialize the sample format to string:
`formats.push(format!("{:?}", WasapiSampleFormat::F32));`
There is no guarantee that this is stable, or matches how the formats are represented in the Yaml config.
Did the LLM you are using suggest this? It's clearly not "nice" Rust, so if an LLM suggested this, I'm curious which one it is.

Wang-Yue:
Thanks for the review. PTAL

> Did the LLM you are using suggest this? It's clearly not "nice" Rust, so if an LLM suggested this, I'm curious which one it is.

I have no knowledge of Rust, so I rely on Claude Opus 3.6 (and sometimes Sonnet 4.6) for all the above code. I believe the `fmt` conversion was due to its recent degradation of reasoning capabilities.

HEnquist:
> I have no knowledge of Rust, so I rely on Claude Opus 3.6 (and sometimes Sonnet 4.6) for all the above code. I believe the `fmt` conversion was due to its recent degradation of reasoning capabilities.

Ok, I mostly use GPT 5.4 and occasionally Claude Opus 4.6. I find these latest models are significantly more reliable at writing high quality code than the previous generation. 

I tested Wasapi and Asio, and both work nicely (after fixing a small error, pushed to your branch). The Wasapi probing is quite slow (which is expected), so I experimented a bit with ways to limit the number of combinations to probe. I came up with a smarter procedure, which seems to work well, but I have not tested it extensively yet. Without this, probing the built-in output of the laptop I'm working on now takes several seconds. With this fix, it's a fraction of a second. The probe results are the same.

Procedure (summarized by GPT 5.4): 

Exclusive mode starts with a fixed safety ceiling for channels, but it tries to shrink that ceiling before doing the broader probe.

It first probes 48 kHz across the candidate channel counts and records, for each channel count, whether any exclusive-mode sample format works.

If 48 kHz works for at least one channel count, the highest working channel count at 48 kHz becomes the new maximum channel count for all later exclusive probing.

If 48 kHz does not work at all, it performs the same channel-count scan at 44.1 kHz and uses that highest working channel count instead.

The successful reference-rate probe is cached and reused later, so that reference rate is not probed twice for the same channel counts.

After the channel ceiling is reduced, the probe works per channel count and explores sample rates in rate families instead of treating every standard rate as unrelated.

The 44.1 kHz family is anchored at 44.1 kHz.

If 44.1 kHz works for a channel count, it probes upward through 88.2, 176.4, 352.8, and 705.6 kHz.

That upward branch stops at the first failure, based on the assumption that if one step in that progression is unsupported, even higher members are unlikely to work.

It also probes downward through 22.05, 11.025, and 5.512 kHz.

That downward branch also stops at the first failure.

The 48 kHz family is anchored at 48 kHz.

If 48 kHz works for a channel count, it probes upward through 96, 192, 384, and 768 kHz.

That upward branch stops at the first failure.

It also probes downward through 16 and 8 kHz.

That downward branch stops at the first failure.

64 kHz and 32 kHz are treated as oddballs in the 48 kHz family.

They are still probed, but they do not control whether probing continues in the rest of the 48 kHz family.

A failure at 64 kHz or 32 kHz does not stop the higher or lower branch.

For every rate that is tried, the code checks all supported WASAPI exclusive sample formats of interest and only records the rate if at least one format succeeds.

The overall effect is that the code no longer tests every possible combination of:

every channel count

every standard sample rate

every exclusive sample format

Instead, it reduces the channel range first, reuses the reference probe, and then prunes large parts of the rate search space by following family structure and stopping branches early when a device has already shown that the progression is unlikely to continue.

HEnquist:
I have merged these changes, as well as the ws stream branch, and the smarter wasapi probing scheme into the next4.2.0 branch. Thanks for putting up with all my comments :)

PR #470, opened 2026-04-24
Spectrum analyzer

HEnquist:
Add commands for reading and subscribing to spectrum data

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -53,7 +53,7 @@ windows = {version = "0.62.0", features = ["Win32_System_Threading", "Win32_Foun
 asio-sys = { version = "0.2.5", optional = true }
 
 [dependencies]
-serde = { version = "1.0", features = ["derive"] }
+serde = { version = "1.0", features = ["derive", "rc"] }
 yaml_serde = "0.10.3"
 serde_json = "1.0"
 serde_with = "3.12"
--- a/src/alsa_backend/device.rs
+++ b/src/alsa_backend/device.rs
@@ -594,6 +594,7 @@ fn playback_loop_bytes(
                 );
                 //trace!("PB: Avail at chunk rcvd: {:?}", avail_at_chunk_recvd);
                 chunk.update_stats(&mut chunk_stats);
+                crate::push_playback_audio_buffer(&params.playback_status, &chunk);
                 conversion_result =
                     chunk_to_buffer_rawbytes(chunk, &mut buffer, &params.sample_format);
 
@@ -1062,6 +1063,7 @@ fn capture_loop_bytes(
             false,
         );
         chunk.update_stats(&mut chunk_stats);
+        crate::push_capture_audio_buffer(&params.capture_status, &chunk);
         crate::update_capture_signal_status(
             &params.capture_status,
             &chunk_stats,
--- a/src/alsa_backend/threaded_device.rs
+++ b/src/alsa_backend/threaded_device.rs
@@ -1229,6 +1229,7 @@ impl PlaybackDevice for AlsaPlaybackDevice {
                                     }
 
                                     chunk.update_stats(&mut chunk_stats);
+                                    crate::push_playback_audio_buffer(&playback_status, &chunk);
                                     let conversion_result =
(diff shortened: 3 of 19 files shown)


Wang-Yue:
That's exciting change! Previously we had to write our own [^1], which is a lot of code and messy and error prone. Then I hijacked into the library to get the data [^2] and move the FFT to the Rust bridge[^3], which is better but hacky. With your change, Those hacky code is completely gone [^4] and it works great!

Suggestion --- I recommend you modularize your bin.rs and move commonly used code to the library, so we don't need to replicate (see [^5])

[^1]: see https://github.com/Wang-Yue/CamillaDSP-Monitor/blob/c013759f571fe38e290cbdfc5a88017a76b9b319/Sources/CamillaDSPMonitor/Models/FFTSpectrumAnalyzer.swift for the FFT and create another CoreAudio thread in https://github.com/Wang-Yue/CamillaDSP-Monitor/blob/c013759f571fe38e290cbdfc5a88017a76b9b319/Sources/CamillaDSPMonitor/Models/CoreAudioTap.swift for capturing the data, and a ring buffer https://github.com/Wang-Yue/CamillaDSP-Monitor/blob/c013759f571fe38e290cbdfc5a88017a76b9b319/Sources/CamillaDSPMonitor/Models/AudioRingBuffer.swift
[^2]: https://github.com/Wang-Yue/CamillaDSP-Monitor/blob/72ec45446b61b55c526396cc3e527df6fc810664/RustBridge/src/engine.rs#L52-L91
[^3]: https://github.com/Wang-Yue/CamillaDSP-Monitor/blob/72ec45446b61b55c526396cc3e527df6fc810664/RustBridge/src/spectrum.rs
[^4]: https://github.com/Wang-Yue/CamillaDSP-Monitor/commit/0ea0a4d560000a838ea2775860ee06f5996973d1
[^5]: https://github.com/Wang-Yue/CamillaDSP-Monitor/blob/0ea0a4d560000a838ea2775860ee06f5996973d1/RustBridge/src/engine.rs and https://github.com/Wang-Yue/CamillaDSP-Monitor/blob/0ea0a4d560000a838ea2775860ee06f5996973d1/RustBridge/src/lib.rs#L35-L120

HEnquist:
> Suggestion --- I recommend you modularize your bin.rs and move commonly used code to the library, so we don't need to replicate (see [5](#user-content-fn-5-b9a7a0184dc101fba1acbe5f535aac5c))

See https://github.com/HEnquist/camilladsp/pull/471
The current bin - lib split was done a long time ago to make it possible to run benches. At the time I didn't consider that anyone would want to use the library part separately. But it does make sense, and it should then be published on crates.io.

Wang-Yue:
> The current bin - lib split was done a long time ago to make it possible to run benches. At the time I didn't consider that anyone would want to use the library part separately. But it does make sense, and it should then be published on crates.io.

The issues with running cdsp binary are 
- Managing the launch process takes some work. Modern OS sandbox such as accessing the file system to find cdsp binary, clearing external process security requirements, get the host and cdsp program to allow microphone access, are always a lot of work. During app start I also need to ensure all cdsp processes are killed (which is very fragile code) before launching a new one, and when I exit I need to kill the process I created. Sometimes my app crashes during development and leaves zombie cdsp process and I have to manually kill it. I also need to handle lost cdsp connection issue. It is a lot of work to manage a state machine just to make sure we can talk to the cdsp binary. 

- socket has higher cost, not only on the performance side (spectrum may be a large json data), but also on the code complexity side, because we are essentially creating a network application, and we need to do state machine management. This is especially true after subscription is introduced. For instance, If I create a state subscription and a level/spectrum subscription, I find in almost all cases if I stop, the last level/spectrum update is sent after the inactive state update, so I have to filter the former out, or create a fake silent level/spectrum update a few hundred ms after getting the inactive state notification, just to make the level meter view showing no signal.  Race conditions like these are difficult to get right. 

With the library it's way easier and less code to maintain. I don't need to handle the network race conditions, state polling is basically free, and sending level/spectrum as floats over FFI is almost zero cost compared to the computation required. I also dont need to manage the cdsp process lifecycle. The app can just use the same state (running, inactive, etc) of camilladsp and I can trust it is working reliably

Issue #469, opened 2026-04-24
Bug Report: Initial Volume Burst in CamillaDSP


Wang-Yue:
## Description
When setting the volume (via WebSocket or ALSA linking) before a configuration is loaded, CamillaDSP exhibits a brief "full volume burst" when audio starts. The audio begins at the default volume (usually 0 dB) and then ramps down to the desired volume over the configured ramp time.

## Root Cause
The issue lies in how the `Volume` filter is initialized and how volume updates are handled when the processing pipeline is not active.

1.  **Initialization**: When a configuration is loaded, the `Volume` filter is created using `Volume::from_config`. It retrieves the `current_volume` from the shared `ProcessingParameters` and uses it as the starting point for both its internal `current_volume` and `target_volume`.
2.  **Delayed Update**: Commands like `SetVolume` only update the `target_volume` in `ProcessingParameters`. They do not update `current_volume` because, during normal operation, the `Volume` filter is responsible for updating the `current_volume` as it ramps.
3.  **The Mismatch**: If `SetVolume` is called while the state is `Inactive` or `Starting`, there is no active `Volume` filter to perform the ramp. `ProcessingParameters` ends up with a new `target_volume` but keeps the old `current_volume` (defaulting to 0 dB).
4.  **The Burst**: When the pipeline finally starts, the new `Volume` filter sees the mismatch between its initial state (0 dB) and the `target_volume` in `ProcessingParameters` (-40 dB, for example). It correctly starts a ramp, but since its starting point was 0 dB, the first few chunks of audio are played at or near full volume.

## Affected Commands
*   `SetVolume` / `AdjustVolume`
*   `SetFaderVolume` / `AdjustFaderVolume`
*   ALSA Volume Linking (startup synchronization)

## Analysis of `SetMute`
`SetMute` is **not** affected. The `Volume` filter's constructor explicitly checks the mute state and, if true, sets the initial gain to zero (-100 dB) immediately, overriding any volume values.

---

## Proposed Fix (Patch)

The fix is to synchronize `current_volume` with `target_volume` in `ProcessingParameters` whenever a volume change is requested while the processing pipeline is not active (`Inactive` or `Starting` states).

```patch
--- a/src/bin.rs
+++ b/src/bin.rs
@@ -428,6 +428,10 @@
                         StatusMessage::SetVolume(vol) => {
                             debug!("SetVolume message to  {vol} dB received");
                             status_structs.processing.set_target_volume(0, vol);
[...]

PR #474, opened 2026-04-27
Update volume handling to set current volume based on processing state

Wang-Yue:
Fix https://github.com/HEnquist/camilladsp/issues/469


HEnquist:
See https://github.com/HEnquist/camilladsp/pull/475 for an alternative safer approach

PR #475, opened 2026-04-27
Check if volume filters missed a vol change

### LINKED ISSUE #469: Bug Report: Initial Volume Burst in CamillaDSP

HEnquist:
Fix: skip volume ramp when processing was stopped during the ramp window
Fixes #469.

When processing is inactive and a volume change arrives, the Volume filter had no way to know that audio wasn't playing when the ramp should have started. On the next processing run it would ramp from the previous volume to the new target, even though no audio had played at the old level — causing an audible sweep that the user didn't expect.

Approach
Rather than patching every call site that sets target_volume (websocket handlers, engine, ALSA backend) to also check processing state and mirror the write to current_volume, the fix lives entirely inside the Volume filter itself.

ProcessingParameters::set_target_volume now records a timestamp (nanoseconds since a program-lifetime epoch, stored as an AtomicU64 per fader). When Volume::prepare_processing detects a target change, it checks whether the request is stale: if more than 1.5× the chunk duration has elapsed since the volume was set without the filter having processed any audio, the ramp is skipped and the new volume is applied immediately.

This makes the fix self-contained and foolproof — it handles every code path that sets volume (including the ALSA hardware volume callback and any future additions) without requiring each one to reason about processing state.

What changes
src/lib.rs: ProcessingParameters gains a target_volume_set_at field ([AtomicU64; NUM_FADERS]), updated atomically on every set_target_volume call. A shared nanos_since_epoch() helper (backed by a lazy OnceLock<Instant>) provides the timestamps.
src/filters/basicfilters.rs: Volume gains stale_ramp_threshold_ns (1.5 × chunk duration, computed once in new()). In prepare_processing, a stale request bypasses the ramp and snaps to the new target directly.

### DIFF
--- a/src/filters/basicfilters.rs
+++ b/src/filters/basicfilters.rs
@@ -31,6 +31,7 @@ use crate::NewValue;
 use crate::PrcFmt;
 use crate::ProcessingParameters;
 use crate::Res;
+use crate::nanos_since_epoch;
 use crate::utils::decibels::gain_from_value;
 
 #[derive(Clone, Debug)]
@@ -60,6 +61,7 @@ pub struct Volume {
     processing_params: Arc<ProcessingParameters>,
     fader: usize,
     volume_limit: f32,
+    stale_ramp_threshold_ns: u64,
 }
 
 impl Volume {
@@ -78,6 +80,7 @@ impl Volume {
         let name = name.to_string();
         let ramptime_in_chunks =
             (ramp_time_ms / (1000.0 * chunksize as f32 / samplerate as f32)).round() as usize;
+        let stale_ramp_threshold_ns = 1_500_000_000u64 * chunksize as u64 / samplerate as u64;
         let current_volume_with_mute = if mute { -100.0 } else { current_volume };
         let target_linear_gain = if mute {
             0.0
@@ -99,6 +102,7 @@ impl Volume {
             processing_params,
             fader,
             volume_limit: limit,
+            stale_ramp_threshold_ns,
         }
     }
 
@@ -156,7 +160,10 @@ impl Volume {
 
         // Volume setting changed
         if (target_volume - self.target_volume).abs() > 0.01 || self.mute != shared_mute {
-            if self.ramptime_in_chunks > 0 {
+            let set_at = self.processing_params.target_volume_set_at(self.fader);
+            let ramp_is_stale =
+                nanos_since_epoch().saturating_sub(set_at) > self.stale_ramp_threshold_ns;
+            if self.ramptime_in_chunks > 0 && !ramp_is_stale {
                 trace!(
                     "starting ramp: {} -> {}, mute: {}",
                     self.current_volume, target_volume, shared_mute
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -383,6 +383,13 @@ impl ProcessingParameters {
         f32::from_bits(self.current_volume[fader].load(Ordering::Relaxed))
     }
 
+    pub fn sync_volumes_to_target(&self) {
+        for fader in 0..Self::NUM_FADERS {
+            let target = self.target_volume(fader);
+            self.set_current_volume(fader, target);
+        }
+    }
+
     pub fn set_current_volume(&self, fader: usize, current: f32) {
         self.current_volume[fader].store(current.to_bits(), Ordering::Relaxed)
     }
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -68,9 +68,10 @@ use serde::Serialize;
 use std::error;
 use std::fmt;
 use std::sync::{
-    Arc,
-    atomic::{AtomicBool, AtomicU32, Ordering},
+    Arc, OnceLock,
+    atomic::{AtomicBool, AtomicU32, AtomicU64, Ordering},
 };
+use std::time::Instant;
 
 pub static SHUTDOWN_REQUESTED: AtomicBool = AtomicBool::new(false);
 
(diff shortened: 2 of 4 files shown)


Wang-Yue:
The fix is not correct --- before the config is read, volume filter is not created.  I recommend you to test with the python script I wrote in https://github.com/HEnquist/camilladsp/issues/469#issuecomment-4326945377 and here's the output with your commit:
```
Starting CamillaDSP...
Connecting to WebSocket...
Setting volume to -100 dB...
SetVolume reply: {'SetVolume': {'result': 'Ok'}}
Loading config...
SetConfig reply: {'SetConfig': {'result': 'Ok'}}
Reading audio data from stdout...
Read 40960 bytes.
Max sample value: 32530
Max value in first chunk: 32530
First 100 samples of first chunk:
(0, 0, 4649, 4649, 9199, 9199, 13556, 13556, 17633, 17633, 21347, 21347, 24624, 24624, 27398, 27398, 29614, 29614, 31227, 31227, 32205, 32205, 32530, 32530, 32195, 32195, 31209, 31209, 29592, 29592, 27378, 27378, 24613, 24613, 21353, 21353, 17664, 17664, 13623, 13623, 9311, 9311, 4816, 4816, 230, 230, -4354, -4354, -8845, -8845, -13151, -13151, -17184, -17184, -20865, -20865, -24118, -24118, -26878, -26878, -29090, -29090, -30710, -30710, -31706, -31706, -32058, -32058, -31761, -31761, -30821, -30821, -29258, -29258, -27105, -27105, -24405, -24405, -21215, -21215, -17599, -17599, -13632, -13632, -9393, -9393, -4971, -4971, -453, -453, 4067, 4067, 8499, 8499, 12754, 12754, 16745, 16745, 20391, 20391)
BUG REPRODUCED: Detected high amplitude signal!
Stopping CamillaDSP...
```

HEnquist:
Updated, this now passes the test script.

Wang-Yue:
I tested. it works with just `self.status_structs.processing.set_target_volume(0, volume);` call, without `set_current_volume`.

HEnquist:
> I tested. it works with just `self.status_structs.processing.set_target_volume(0, volume);` call, without `set_current_volume`.

Where do you mean?

Wang-Yue:
I mean with your change,  in my rust bridge, I can use 
```rust
    pub fn set_volume(&self, fader: u32, volume: f32) {
        self.status_structs
            .processing
            .set_target_volume(fader as usize, volume);
    }
```
for the volume change now. without adding additional 
```rust
        let state = self.status_structs.capture.read().state;
        if state == ProcessingState::Inactive || state == ProcessingState::Starting {
            self.status_structs.processing.set_current_volume(0, volume);
        }
```

so your change did fix the issue.

HEnquist:
Got it, thanks! I added a test similar to your python script as well, to make sure the issue doesn't come back.

PR #477, opened 2026-06-18
FileWriter processor

brainrake:
It's a processor that writes to a file.
Shares code with `FilePlaybackDevice`. 

Todo:
- Joining all the threads introduces a lot of maintenance burden, looking for a less intrusive solution.
- Doesn't support a channel list. All channels are saved.
- `%timestamp%` support
- bugs

### DIFF
--- a/README.md
+++ b/README.md
@@ -2487,33 +2487,6 @@ pipeline:
   * `monitor_channels`: a list of channels used when estimating the loudness. Optional, defaults to all channels.
   * `process_channels`: a list of channels to be gated. Optional, defaults to all channels.
 
-### FileWriter
-The "FileWriter" processor copies the samples passing through the pipeline to a separate file writer buffer.
-A writer thread drains this buffer and writes the data to a file, so file writing does not modify the audio stream.
-If the file writer cannot keep up and the ring buffer becomes full, new recorded bytes are dropped and processing continues.
-
-Example:
-```
-processors:
-  recorder:
-    type: FileWriter
-    parameters:
-      channels: 2
-      filename: "recording"
-      format: S32_LE
-      wav_header: false (*)
-
-pipeline:
-  - type: Processor
-    name: recorder
-```
-
-  Parameters:
-  * `channels`: number of channels, must match the number of channels of the pipeline where the recorder is inserted.
-  * `filename`: output file name.
-  * `format`: file sample format, one of `S16_LE`, `S24_4_RJ_LE`, `S24_4_LJ_LE`, `S24_3_LE`, `S32_LE`, `F32_LE`, `F64_LE`.
-  * `wav_header`: write a wav header before the samples. Optional, defaults to `false`. Wav files do not support `S24_4_RJ_LE`.
-
 ### RACE
 The "RACE" processor implements the recursive part of the
 [Recursive Ambiophonic Crosstalk Elimination (RACE)](http://www.filmaker.com/papers/RGRM-RACE_rev.pdf) algorithm.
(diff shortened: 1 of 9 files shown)


HEnquist:
Nice progress, and I know it's still a draft, so this is about direction rather than nitpicks.

One thing is a hard no for me: allocations in the real-time path. `process_chunk` does `input.waveforms.clone()` on every chunk, which heap-allocates in the audio thread. That can't go in.

It traces back to the transport. Sending a whole `AudioChunk` over the channel means you have to own it, hence the clone, and it's also why channel selection is awkward (your TODO note about saving all channels). Both come from the same choice.

What I had in mind solves both at once: interleave the selected channels into a preallocated ring buffer on the audio thread. No allocation, since you copy into existing memory, and channel selection is just which channels you read in the interleave loop. The writer thread then just drains bytes and writes them.

For the ring buffer, use the `ringbuf` crate we already depend on, the `HeapRb` producer/consumer split, same as the WASAPI backend does in `src/wasapi_backend/device.rs`. No need to pull in anything new.

The rest is good and worth keeping: the shared wav header / conversion / finalize code, and the `shutdown()` + reaper handling for clean teardown. It's really just the transport and writer loop that need to change.

brainrake:
I had no allocations but then I rewrote it to be more like FilePlaybackDevice :):) Almost exactly as you describe it.

HEnquist:
Sharing code is good, but not allocating in the RT thread is better! Go back to your no-alloc ring buffer please :)

brainrake:
Also spawning, and joining (finished) threads in the processing thread (during reconfigure).

brainrake:
Meh. Will start a new one with a better design.

HEnquist:
Posting a design I hashed out with Claude, so you have it before getting far into the rewrite.

## FileWriter design notes

A few notes to keep the next version small. The key realization is that we already write a **streaming WAV header**: `write_wav_header` sets the RIFF and data sizes to `u32::MAX` and nothing ever seeks back to patch them (same as `FilePlaybackDevice`). So there is **no finalize step**, and a file cut off mid-write is still a valid WAV, just shorter.

That single fact removes most of the complexity from the earlier attempt.

### Transport

- A preallocated lock-free ring buffer, the `ringbuf` crate we already use (`HeapRb`, producer/consumer split, same as the WASAPI backend in `src/wasapi_backend/device.rs`).
- `process_chunk` interleaves the selected channels straight into the ring buffer. No `AudioChunk` clone, no per-chunk allocation. If the buffer is full, drop and move on (best effort).
- Channel selection is just which channels you read in the interleave loop, so it costs nothing and is easy to add.
- Fixed buffer size is fine, no need for a config knob.

Don't reuse the `FilePlaybackDevice` channel-of-`AudioChunk` transport. Owning a chunk to send it is exactly what forced the allocation last time.

### Writer thread

- Spawn one writer thread per FileWriter in `from_config`. It owns the ring buffer consumer and the file: write the header once, then loop draining the ring buffer to disk.
- You can either interleave to raw bytes on the audio thread and have the writer just `write_all`, or push samples and convert in the writer. Either is fine. Reuse `write_wav_header` and the conversion helpers in `src/utils/conversions.rs`.

### Lifecycle (the simple part)

Because there's no finalize, you never need to join for correctness:

- On `Drop` (pipeline rebuild), just drop the producer. The writer sees the ring buffer closed, drains what's left, flushes, and exits on its own. Drop/detach the `JoinHandle`.
- No reaper, no `shutdown()` trait method, no thread juggling in the processing loop.
- Spawning a thread during a reconfigure is fine, that path already rebuilds the whole pipeline between chunks, it's not the realtime hot path. Only `process_chunk` needs to be allocation-free.
- At process exit you might lose the last fraction of a second still in the buffer. That's acceptable for a best-effort diagnostic tap, and the file stays valid.

That's the whole thing: ring buffer in, writer thread drains to disk, drop to stop.

PR #478, opened 2026-06-18
tidy bin: move signals into module, name threads

kvc0:
Hi @HEnquist! I've set up CamillaDSP as a little DAC appliance to go with my KVM that switches between macos, windows, and linux. I want my DSP to follow across operating systems, and I don't want to figure out all 3 os's :-). It's a pretty nice setup; I've even got a dynamic config & output swap for headphones vs speakers output.

Anyway, CamillaDSP is an incredible piece of software. I could not have reasonably made this desktop DAC without it. I'm a rust developer, and I wonder if you might be open to some janitorial PR's? I don't rely on ai for my open source pr's, so I'm hopeful these efforts won't be seen as extractive. I just like this software and I would love to see some edges (like thread names and file cohesion, maybe workspace organization eventually) cleaned up a little.

I figure to make the most of your time I can dump a few low hanging fruit cleanups in this PR. Each commit is focused on one change, and the only behavior I'm changing is to add names to threads (so my htop looks more friendly)

### DIFF
--- a/benches/conversions.rs
+++ b/benches/conversions.rs
@@ -1,7 +1,5 @@
-extern crate criterion;
 use camillalib::utils::stash::{recycle_chunk, vec_from_stash};
 use criterion::{Criterion, criterion_group, criterion_main};
-extern crate camillalib;
 
 use camillalib::audiochunk::AudioChunk;
 use camillalib::config::BinarySampleFormat;
--- a/benches/fftconv_kernels.rs
+++ b/benches/fftconv_kernels.rs
@@ -1,10 +1,6 @@
 //! Micro-benchmarks for fftconv complex-multiply kernels.
 //! Compares scalar with AVX+FMA on x86_64 and NEON on aarch64.
 
-extern crate camillalib;
-extern crate criterion;
-extern crate num_complex;
-
 use criterion::{BatchSize, BenchmarkId, Criterion, Throughput, criterion_group, criterion_main};
 use num_complex::Complex;
 use std::hint::black_box;
--- a/benches/filters.rs
+++ b/benches/filters.rs
@@ -1,6 +1,4 @@
-extern crate criterion;
 use criterion::{Bencher, BenchmarkId, Criterion, criterion_group, criterion_main};
-extern crate camillalib;
 
 use camillalib::PrcFmt;
 use camillalib::filters::Filter;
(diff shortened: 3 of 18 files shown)


HEnquist:
Hi @kvc0,

Thanks a lot for this, it is exactly the kind of cleanup I like to see, and the KVM/appliance build sounds great. Glad CamillaDSP is pulling its weight there. Is there anywhere I can read more about that setup? It sounds like a fun project.

First a note on timing, since it explains most of what follows. `master` always matches the latest release, so it is usually behind. The real development happens on a branch named after the next version, currently `next4.2.0`. That branch already went through a fairly big restructuring, so several of your changes have either already landed there or no longer apply cleanly.

What is already done on `next4.2.0`:

- Thread naming. `AlsaPlayback`, `AlsaCapture`, `SignalGenerator` and the inner threads all get names now.
- Signal handling has moved out of `bin.rs` and lives in `engine.rs`.

On the ALSA deduplication: I can see why the two backends look like copy/paste asking to be merged, but the duplication is intentional for now. There is a new threaded ALSA backend coming up alongside the existing one, and I want to keep them mostly independent until the new one takes over and the old one is dropped. Merging them now would couple the two paths, so I am keeping `open_pcm`, the mutex and `DeviceBufferManager` separate for the time being.

The `extern crate` removal was still useful and applied cleanly, so I pulled it into `next4.2.0` in c925bdf. Thanks for the nudge.

Finally, since the `master` confusion bites every new contributor, I added a short note about the branch strategy to the README, in ad98507. Short version: ask which `next*` branch to target before starting, and you will not end up rebasing onto a restructure like this one.

Cheers,
Henrik

kvc0:
Thank you @HEnquist ! I’ll write up a blog and shoot you a note at some point about the build. The story isn’t quite complete yet: I’m building small desktop speaker cabinets right now in my garage with DSP in mind. They’ll need a linkwitz transform.

Thanks for clarifying the development location - I will check in the future. Hopefully I’ll have time to inspect 4.2.0 and make another change next week! I’ll be more careful about deduplication with a little more familiarity with your workflow.

kvc0:
> Glad CamillaDSP is pulling its weight there. Is there anywhere I can read more about that setup? It sounds like a fun project.

@HEnquist my project is mostly complete for now. I wrote up a post about it, if you still wanted to read more!

https://kvc0.com/posts/2026/desktopaudio/

PR #480, opened 2026-06-20
FileWriter processor

brainrake:
Processor that saves audio to a file.

### DIFF
--- a/README.md
+++ b/README.md
@@ -2627,6 +2627,7 @@ pipeline:
 ### FileWriter
 The "FileWriter" processor writes the audio to a file.
 This is useful for recording the (un)processed audio, or for capturing intermediate stages of the pipeline for analysis.
+When exiting the process, the last few chunks may not be written to the file.
 
 Example:
 ```yml
@@ -2635,6 +2636,7 @@ processors:
     type: FileWriter
     parameters:
       channels: 2
+      process_channels: [0, 1] (*)
       filename: /tmp/recording.wav
       format: S16_LE
       wav_header: true (*)
@@ -2642,6 +2644,7 @@ processors:
 
   Parameters:
   * `channels`: number of channels, must match the number of channels of the pipeline where the FileWriter is inserted.
+  * `process_channels`: a list of channels to write to the file. Optional, defaults to all channels. The channels are written in the specified order.
   * `filename`: path to the output file. The file is overwritten if it already exists.
   * `format`: sample format of the output file. One of `S16_LE`, `S24_3_LE`, `S24_4_RJ_LE`, `S24_4_LJ_LE`, `S32_LE`, `F32_LE`, `F64_LE`.
   * `wav_header`: prepend a WAV header to the file. Optional, defaults to `false`. When `true`, the format must not be `S24_4_RJ_LE`.
--- a/README.md
+++ b/README.md
@@ -2625,8 +2625,8 @@ pipeline:
 ```
 
 ### FileWriter
-The "FileWriter" processor writes the audio to a file.
-This is useful for recording the (un)processed audio, or for capturing intermediate stages of the pipeline for analysis.
+The "FileWriter" processor writes raw audio to a file.
+This is useful for recording (un)processed audio, or for capturing intermediate stages of the pipeline for analysis.
 When exiting the process, the last few chunks may not be written to the file.
 
 Example:
@@ -2637,17 +2637,15 @@ processors:
     parameters:
       channels: 2
       process_channels: [0, 1] (*)
-      filename: /tmp/recording.wav
+      filename: /tmp/capture
       format: S16_LE
-      wav_header: true (*)
 ```
 
   Parameters:
   * `channels`: number of channels, must match the number of channels of the pipeline where the FileWriter is inserted.
   * `process_channels`: a list of channels to write to the file. Optional, defaults to all channels. The channels are written in the specified order.
-  * `filename`: path to the output file. The file is overwritten if it already exists.
+  * `filename`: path to the output file. The file is never overwritten: a zero-padded counter is appended to the filename to claim a unique name, e.g. with `filename: /tmp/capture` the files written are `/tmp/capture.000`, `/tmp/capture.001`, and so on. The counter starts one higher than the highest numbered file already present.
   * `format`: sample format of the output file. One of `S16_LE`, `S24_3_LE`, `S24_4_RJ_LE`, `S24_4_LJ_LE`, `S32_LE`, `F32_LE`, `F64_LE`.
-  * `wav_header`: prepend a WAV header to the file. Optional, defaults to `false`. When `true`, the format must not be `S24_4_RJ_LE`.
 
 ## Pipeline
 The pipeline section defines the processing steps between input and output.
--- a/README.md
+++ b/README.md
@@ -136,6 +136,7 @@ and the Mozilla Public License Version 2.0:
    - **[NoiseGate](#noise-gate)**
    - **[Lookahead limiter](#lookahead-limiter-processor)**
    - **[RACE](#race)**
+   - **[FileWriter](#filewriter)**
 - **[Pipeline](#pipeline)**
    - **[Filter step](#filter-step)**
    - **[Mixer and Processor step](#mixer-and-processor-step)**
@@ -2622,6 +2623,28 @@ pipeline:
   - type: Mixer
     name: 6to2
 ```
+
+### FileWriter
+The "FileWriter" processor writes the audio to a file.
+This is useful for recording the (un)processed audio, or for capturing intermediate stages of the pipeline for analysis.
+
+Example:
+```yml
+processors:
+  recorder:
+    type: FileWriter
+    parameters:
+      channels: 2
+      filename: /tmp/recording.wav
+      format: S16_LE
+      wav_header: true (*)
+```
+
+  Parameters:
+  * `channels`: number of channels, must match the number of channels of the pipeline where the FileWriter is inserted.
+  * `filename`: path to the output file. The file is overwritten if it already exists.
+  * `format`: sample format of the output file. One of `S16_LE`, `S24_3_LE`, `S24_4_RJ_LE`, `S24_4_LJ_LE`, `S32_LE`, `F32_LE`, `F64_LE`.
+  * `wav_header`: prepend a WAV header to the file. Optional, defaults to `false`. When `true`, the format must not be `S24_4_RJ_LE`.
 
 ## Pipeline
 The pipeline section defines the processing steps between input and output.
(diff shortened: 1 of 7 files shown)


brainrake:
Todo: 
- [x] channel selection
- ~`$timestamp$` and other variable support~
- [x] buffer sizing?
- [ ] testing

HEnquist:
Design looks great so far, nice work. Ping me once it's ready and I'll go through it properly.

HEnquist:
Hi Márton, hope you're well. Two heads-ups. First, there won't be a 4.2 release, so the target branch is now `next5` rather than `next4.2.0`. Second, #491 added RF64 support and proper size patching for wav output in the File backend, so this conflicts now, and the FileWriter wav path should probably reuse that instead of the streaming header with placeholder sizes.

Are you still up for finishing it? No rush from my side, but if you're busy I'd like to pick it up. In that case I'd branch off your commit so it stays attributed to you, and finish the remaining items there.

HEnquist:
Hi again, I looked a bit closer at the wav side and want to walk back part of what I said.

Reusing the RF64 and size patching from #491 doesn't fit here. That writer only gets correct sizes if `finalize` runs, and until then the header says zero length, not unknown length. A recording that gets killed mid-write then looks empty, even though the samples are on disk. RF64 has the same problem, its ds64 fields also start at zero. The streaming header you already use has no finalize step, so a file that gets cut off is still valid, just shorter. Keep that as it is. Sharing the header writing and the conversion helpers still makes sense, just not the seekable writer.

The catch with the streaming header is size. It passes 4 GB without complaining, but the size field is maxed out, so any reader that trusts it stops there. 8 channels of 32-bit at 48 kHz reaches that in about 45 minutes.

That's fine for a diagnostic tool. Grab a few seconds or minutes from any point in the pipeline, best effort, and a valid file even if it gets cut off. That's a simple addition, pretty much what you already made, plus channel selection and the timestamp token.

A proper recorder is a different project. Correct lengths for long sessions means RF64 with the sizes patched while recording, and on top of that dropped sample accounting, disk full handling, and sooner or later file splitting and start/stop control. I maintain this alone, so that is not something I can take on. The diagnostic version is the only feasible one here.

Minor: tokens in CamillaDSP use `$name$`, so `$timestamp$` rather than `%timestamp%`.

The offer from last time stands. If you're busy I can finish it off on a branch from your commit.

brainrake:
Thank you! I want to test this soon.

brainrake:
Successfully tested this to record multiple days worth of audio. There are some minor problems remaining, I will follow up with patches.
- Without timestamps, the file is overwritten at pipeline rebuild or buffer underrun
- I added timestamp support, will push patch
- When unable to start (eg bad device settings), with timestamp enabled, many empty or header-only files are created.

brainrake:
I tried to come up with a minimal complexity solution that prevents unwanted overwrites and litter. Landed on an automatic, non-configurable counter (dropped timestamp, fs has it anyway). It scans directory to always start at largest existing counter plus one, should be fine on non-audio thread. 
Also removed header, size was incorrect in any case. Not 100% sure about this one.

brainrake:
Will do some more testing.
Any guidance on buffer size? Currently `ring_size = write_channels * samplerate.max(MIN_CHUNKS * chunksize * write_channels)`. So at least one second and at least 4 chunks.

HEnquist:
Great to see this moving again, and the multi-day test is a much better shakedown than I would have managed.

**On the numbered files:** I would rather drop them and bring back `$timestamp$` instead. The overwriting you hit is a bug elsewhere, and the automatic counter is treating the symptom.

The new `config_diff` arm is guarded on `params != current_proc`, so when the FileWriter is *unchanged* the guard fails and it falls through to the `_` arm, which returns `ConfigChange::Pipeline` as well. Every config reload therefore rebuilds the pipeline whenever a FileWriter is in the config, restarts the writer thread and truncates the file. Flipping the guard to `params == current_proc` with an empty body should fix it.

The other half, the empty and header-only files after a failed start, is already solved by the lazy open you added. No data means no file, so a bad device config now leaves nothing behind. That fix stands on its own and I like it. It is also what makes the timestamp workable now, since that litter is exactly what killed the idea in June.

My problem with the automatic counter is that it can never be turned off. A FileWriter left in a config on a machine that runs around the clock then fills the disk with no upper bound, and losing a scratch file is the better failure of the two. A `$timestamp$` has the same growth, but only if you ask for it, and then it is what you wanted. Plain `filename` stays predictable, which matters when you want to script something against the output.

A few notes if you pick it up. `chrono` is already a dependency, so nothing new is needed. Resolve the timestamp once in `replace_tokens_in_config` and pass it down rather than calling `now()` inside `replace_tokens`, which runs once per filename and would drift if it crosses a second. Note that `replace_tokens_in_config` only touches coefficient filenames and pipeline step names today, never processor parameters, so the FileWriter filename needs wiring in regardless. I would scope `$timestamp$` to the FileWriter filename only and document it there rather than in the general token section, since it is meaningless in a filter name or a coefficient path. Something like `%Y%m%d_%H%M%S` in local time works, avoid colons so Windows is happy.

Since substitution happens at config load, one load gives one file, and a device that fails and retries reuses the same name instead of spraying new ones. The resolved path is what ends up in the stored config, so `GetConfig` will show the expanded filename, same as `$samplerate$` does today.

**On the header:** the `u32::MAX` in the size fields is the streaming convention for "length unknown, read to EOF", not a broken value waiting to be patched. We already do exactly this for the `Stdout` playback device and the README documents it. I would like it back, since without it the file is headerless raw and samplerate, channel count and format have to be re-entered by hand in whatever opens it, which is the metadata a diagnostic capture most wants to carry.

**Buffer size:** one second and at least 4 chunks is exactly right, the code just does not quite say that. In `write_channels * samplerate.max(MIN_CHUNKS * chunksize * write_channels)` the left side of the `max` is frames, the right side is already samples, and then the whole thing is multiplied by `write_channels` again. At 8 channels, 44100 and chunksize 16384 that comes to 4.2 M samples, around 33 MB, where 4 chunks is 524 k. It should be `write_channels * samplerate.max(MIN_CHUNKS * chunksize)`. At 2 channels and chunksize 1024 both forms agree, which is why testing would not show it.

Smaller things:

- The docs should recommend `F32_LE` for diagnostic use. The integer formats clip at full scale, so anything hotter than 0 dBFS at that point in the pipeline gets saturated on the way to disk, which is often the exact thing you were trying to look at. I checked, `F32_LE` passes 2.5 and -3.0 through untouched with zero clipped samples, `S16_LE` saturates both.
- Same root cause: the conversion helper still logs "Clipping detected" from the writer thread, so an integer FileWriter prints warnings that read as playback clipping. Either a quiet variant or the processor name in the message would fix it.
- `self.warned = false` resets after every successful push, so an overrun hitting every other chunk warns every other chunk rather than once per episode.
- The README note about the last chunks possibly not being written at exit is fine by me, no need to chase that one.

Went through this with a bit of Claude help. I have not run it on real hardware, so apart from the conversion check above it is all from reading the code. Shout if any of it looks off.

brainrake:
Restored wav_header flag.

Fixes:
- config_diff: unchanged FileWriter no longer forces a pipeline rebuild
- ring buffer size formula corrected
- README recommends F32_LE for diagnostic use

Looks like I need to review a bit more before pushing.
Continuing.

brainrake:
- no clipping log from FileWriter
- ring buffer overrun warning resets when at least half of the ring buffer is vacant. neither full nor empty work well (it can flip flop near full or not completely drain and never reset). so half.

brainrake:
Turns out I introduced the pipeline rebuild bug after testing. During testing, I was not triggering it on every parameter change. It's a red herring.

I got into situations where a new FileWriter was started, eg. systemd service reboot loop when device was not available, or adding an output to a mixer section. Without timestamp/counter these cause the output file to be overwritten.

There is another problem, a race condition that results from dropping FileWriter thread handles: the new thread may open and truncate the file before the old thread finishes writing, so the old thread can still write a few samples near the beginning of the new file, inserting glitches. Minor but annoying.

My thinking: it's never a good idea to truncate the file during operation, so we shouldn't allow it.

Between counter and timestamp, either or both could work. Counter is more deterministic and simpler to implement. Counter is also straight forward to do with zero configuration, avoiding the template system. Timestamp is somewhat more complex, and the data is already included in the file system. Timestamp would probably also need to implement a retry loop with incrementing seconds.

Regarding space: a running instance will consume space unboundedly in any case. Using pipeline rebuilds to truncate the file for space seems very unusual. Better use something like mount quota or systemd quota. I use a separate partition for data.

So that's why I preferred an automatic counter. No data loss, no race condition, deterministic, simple, zero-config.

Here's my decision tree atm: one of:
- zero-config counter
- add %timestamp% and %counter%, if neither present and file already exists, one of:
   - data loss and corruption: warn and truncate existing file and maybe glitch it (nah)
   - denial of service: warn and stop (meh)
   - config bool to switch between the above (meh)
   - fallback: counter (ok)
- something else?

I'll leave it up to you. Cheers!

HEnquist:
Hi Márton, thanks for digging into this, and good catch on the race.

I've been looking at it and it's trickier than I thought. Both of us have been treating the naming as the problem, but the real issue is that the writer thread dies on every pipeline rebuild. I want to try a couple of things there before picking between counter and timestamp.

So hang on a bit, and don't build either option in the meantime. No point spending your time on something I might change my mind about.

Unrelated heads-up: next5 dropped the Result from Processor::process_chunk on 1 September, so the branch doesn't compile against it any more. One line in filewriter.rs.

brainrake:
Thank you for your continued effort in shepherding this feature.
I agree, the pain point is that writer threads are detached and so not able to closed and that they do not survive pipeline rebuilds (keeping on handle to the buffer would be enough here). A new file is unnecessary on rebuild. Naming comes as a secondary problem to this.
Owerwriting is still a problem on service restart.

HEnquist:
Thanks for digging into the overwriting. You were right, the real problem is the race when the pipeline is rebuilt, not the `config_diff` thing.

I tried keeping the writer thread alive across rebuilds instead, and it works well: https://github.com/HEnquist/camilladsp/tree/filewriter_pool

- An unchanged FileWriter keeps its writer and file across reloads, and a changed one waits for the old writer to finish before opening the file. That fixes the race, and GUI edits no longer start new files.
- Writers are joined at shutdown, so the last chunks get written.
- Two FileWriters on the same file, or a missing output directory, is now a config error.
- The ring buffer is replaced by normal AudioChunks from the stash. That let the writer reuse the existing conversion and recycling code, and partial chunks can be written now.
- The counter is gone. It was really protecting against the race and against two writers sharing a file, and both are handled now. An opt-in `$timestamp$` in the filename gives a new file per writer instead.

It's your branch with next5 merged in. If it looks ok to you, I'll open it as a PR against your branch so it all stays in #480.

PR #482, opened 2026-06-26
add support for multiple pipelines

kvc0:
# Context

I'm using camilladsp to directly front as an alsa device to multiple KVM'd computers. It works great for audio output, but I have had to resort to alsaloop for getting the mic back through to the host.

```
host A,B,C -> kvm -> pi
                   /   \
         camilladsp   alsaloop
          dac2 pro     volt1
          speakers      mic
```

Alsaloop has problems when I try to also use that secondary sound card (volt1) as a headphones output. It goes into buffer overrun and it causes the volt1 device to restart every 1-3 seconds.

```
host A,B,C -> kvm -> pi
                   /
         camilladsp
    dac2 pro     volt1
    speakers      mic
```

This change makes it possible to do the above. I have tested by running it on my raspberry pi, and it runs both the previous config and the new config with 2 device groups and 2 pipelines.

# Important stuff to note

I haven't made web ui support for this yet. I think there are probably a few different ways it could be exposed, and I'd like to contribute the capability, even if it's relegated to "advanced" users for a time. I've stubbed the websocket config references to group(0), but **it doesn't seem to work with CamillaGUI 4.1.0**. I'm not sure if this is a next4.2.0 known thing or a new regression with this change.

# Reviewing

I recommend looking at `exampleconfigs/multi_pipeline.yml` to get an idea what you're looking for. Changes are in `devices` and `pipelines`. All the other examples should still work and all the tests (that I could run) still work.

Github is rendering the diff pretty uncharitably. I did massage the related code a little to avoid making a mess with device multiplicity. Github is struggling with indentation adjustments, like in validate_config. Local git renders it a little better imo.

* Config gets `DeviceGroups` (supports single/legacy/current format and yaml array)
* new `pipelines` root element, intended to replace `pipeline` eventually. `pipeline` still works, `pipelines` just has the ability to name the device group to which it applies.
* New `RunningPipeline` struct for holding the big bag of channels, thread handles, synchronization, and status variables for each device pipeline. I did not know what needed to be extracted into this before making this change, unfortunately. I think it improves readability but it does conflate a cleanup with a feature.

### DIFF
--- a/Cross.toml
+++ b/Cross.toml
@@ -16,7 +16,11 @@ passthrough = [
 ]
 
 [target.aarch64-unknown-linux-gnu]
-image = "cross/armv8:v1"
+# image = "cross/armv8:v1"
+pre-build = [
+    "dpkg --add-architecture arm64",
+    "apt-get update && apt-get install --assume-yes libasound2-dev:arm64",
+]
 
 [target.aarch64-unknown-linux-gnu.env]
 passthrough = [
--- a/benches/pipeline.rs
+++ b/benches/pipeline.rs
@@ -163,7 +163,7 @@ fn build_pipeline(chunksize: usize, multithreaded: bool, with_conv: bool) -> Pip
     let conf = config::Configuration {
         title: None,
         description: None,
-        devices: config::Devices {
+        devices: config::DeviceGroups(vec![config::Devices {
             samplerate: 48000,
             chunksize,
             queuelimit: None,
@@ -193,7 +193,7 @@ fn build_pipeline(chunksize: usize, multithreaded: bool, with_conv: bool) -> Pip
             volume_limit: None,
             multithreaded: Some(multithreaded),
             worker_threads: None,
-        },
+        }]),
         mixers: Some(mixers),
         filters: Some(filters),
         processors: None,
@@ -216,10 +216,11 @@ fn build_pipeline(chunksize: usize, multithreaded: bool, with_conv: bool) -> Pip
                 bypassed: Some(false),
             }),
         ]),
+        pipelines: None,
     };
 
     let processing_params = Arc::new(ProcessingParameters::new(&[0.0_f32; 5], &[false; 5]));
-    Pipeline::from_config(conf, processing_params)
+    Pipeline::from_config(conf, 0, processing_params)
 }
 
 fn make_chunk(channels: usize, frames: usize) -> AudioChunk {
--- a/exampleconfigs/multi_pipeline.yml
+++ b/exampleconfigs/multi_pipeline.yml
@@ -1,0 +1,82 @@
+---
+# Example: two independent I/O pipelines.
+#
+# `devices` is a list of device groups. Each group is a fully independent
+# capture -> processing -> playback chain with its own sample rate and chunk
+# size, so two soundcards can be processed separately and simultaneously.
+#
+# Each entry in `pipelines` is the processing chain for one device group,
+# selected by its `device_group` index (0 = the first group in `devices`).
+
+devices:
+  # Device group 0: first soundcard, running at 48 kHz.
+  - samplerate: 48000
+    chunksize: 1024
+    capture:
+      type: Alsa
+      channels: 2
+      device: "hw:0"
+    playback:
+      type: Alsa
+      channels: 2
+      device: "hw:0"
+  # Device group 1: second soundcard, running at 44.1 kHz.
+  - samplerate: 44100
+    chunksize: 1024
+    capture:
+      type: Alsa
+      channels: 2
+      device: "hw:1"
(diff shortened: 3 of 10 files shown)


HEnquist:
Hi @kvc0,

Thanks for putting real work into this, and the serde compatibility shim for single-vs-multi device groups is nicely done. But I want to be upfront before you sink more time in.

First, a process note, and I mean this as a heads-up for next time, not a knock on the effort. A change this size, one that reshapes the engine and the config schema, needs to start as an issue or a short discussion about direction, not arrive as a finished PR. Direction is the expensive thing to get wrong, and once a full implementation exists there is a lot of sunk work pulling the decision toward "merge it." For anything architectural, float the idea first and we can sort out whether it fits before you write a line. It saves us both time.

On the substance: multi-pipeline in a single config isn't a road I want to take CamillaDSP down. A few reasons, all pointing the same way:

- **Coupled lifecycles.** The whole appeal of separate devices is independence, but one config means one engine: a single capture or playback error on any group restarts all of them, and they share one global rayon pool. So you get the coupling costs without the independence. Two devices that should be independent are better served by two independent things.
- **Master volume stops meaning anything.** What does the volume control set when there are three pipelines? Same for mute, signal levels, clipped samples, processing load, spectrum. Every scalar in the websocket API either becomes a quiet lie or grows a group index. In this PR it is resolved by only ever exposing group 0, which is why groups beyond the first have volume and mute that nothing can read or set.
- **GUI blast radius.** Every meter, fader, and the config editor currently assumes one pipeline. Going multi means a group selector threaded through the entire UI, and the single-pipeline user (essentially everyone) pays that complexity forever. That is the part you already ran into with CamillaGUI.

The use case itself is good, and you can already get it today by running one CamillaDSP instance per soundcard. That keeps them genuinely independent, which is what you actually want for a mic loopback next to a headphone out.

One thing I'd genuinely welcome as a separate PR: the supervisor cleanup buried in here. Pulling the per-pipeline state into a struct and consolidating the error/teardown arms into one helper is an improvement that stands on its own, no multi-pipeline needed. If you are up for splitting that out against next4.2.0, I'd take it.

HEnquist:
Closing this, since it's the direction I'm turning down rather than anything about the code. The separate cleanup PR offer stands.

kvc0:
I knew there was risk to adding the feature but I wanted to see if you even wanted to do multi-pipeline. It was a measured effort to stop before also reaching into the gui :-)

I can fix the coupled lifecycles, and I can separate the rayon pools. It’s also straightforward to do something like optionally bind device pipelines past 0 to faders or optionally share the master volume (2 sound cards, one for fullrange, one for subwoofers).

Camillagui could appear the same, and have a + button to the right of the device column that adds another device column. If populated, then the pipelines screen could have the single row added to select device group; it could even be presented kind of like the channel selector.

My headphones should move over to the volt1 as well, which would seem to ask for a third instance of camilladsp? It’s a lot to manage, whereas a slim config yml puts all 3 in one simple place.

I have a little http server and actions on my phone to switch between speakers and headphones, and some automation for volume. It’s more to have to monitor and control 3 Camilla websockets than 1.

So I can demonstrate answers to the technical and utility concerns, but I can’t change your will if you don’t want to take camilladsp down this road! Either way, I’m happy to try to keep this functionality alive on my fork, because it has simplified my life 😅

Thank you for the review @HEnquist! I’ll see about extracting the bits you mentioned.

HEnquist:
Looking forward to the cleanup PR.
One more thought, separate from the PR, and you've probably already considered this. It looks a bit like what you actually need is a sound server, something doing dynamic routing across hosts as the KVM switches, more than a DSP engine. If you haven't looked into it, PipeWire with WirePlumber might be worth a look, could fit that role and let you drop alsaloop. CamillaDSP already has a PipeWire backend, so it could still sit in there doing the DSP, as a client in the graph rather than owning the devices.

kvc0:
I considered 3 full stack alternatives over the course of a few weeks. My kvm is solved, and it's a usb switcher. Ultimately I need a usb sound card with flexible DSP internally, with as little latency as possible, as reliable as possible. So I removed unnecessary things until I just had a pi in usb gadget mode, a couple systemd services, and camilladsp. With alsa, it's about as thin as I can make it without making a dkms patch. Actually, :laughing: uh I _did_ also made a dkms patch so I could advertise the device to the host OS's with a vanity name.

The alsaloop caused me problems, so I did this change mostly yesterday. I'll be continuing to use this pr's feature, because it keeps my setup much simpler. Camilladsp does everything I need, and much more, with this yml extension. No extra sound server needed. Camilladsp is sort of a low-level, focused feature set sound server anyway, whether it was intended to be or not :stuck_out_tongue_winking_eye:

PR #483, opened 2026-06-27
factor out EnginePipeline

kvc0:
To tighten up engine::run a little, I pulled out a fail_and_restart
helper and extracted pipeline channel and synchronization bits into
EnginePipeline. I made a tiny, tightly scoped macro to reduce line
noise where fail_and_restart() is called. The macro is intended for
these exact lexical callsites, because of local variable value types
and names.

I left the start_pipeline() function as a module-level function, as
it's more than a constructor / ::new(). I also named the
interactions and kept the fields private. I kept unwrap vs log
semantics unchanged.

Tests pass, file io looks fine - no behavior change was intended.
I tried to lift the bits of https://github.com/HEnquist/camilladsp/pull/482 that made sense to carry over and
rejigger the things that didn't fit as well with the single pipeline
model.

### DIFF
--- a/src/engine.rs
+++ b/src/engine.rs
@@ -16,9 +16,9 @@
 
 use crossbeam_channel::select;
 use parking_lot::{Mutex, RwLockUpgradableReadGuard};
+use std::sync::Arc;
 #[cfg(any(windows, feature = "websocket"))]
 use std::sync::atomic::AtomicBool;
-use std::sync::{Arc, Barrier};
 use std::thread;
 #[cfg(any(windows, feature = "websocket"))]
 use std::time::Duration;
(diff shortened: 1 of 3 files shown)


HEnquist:
This is exactly the cleanup I had in mind, thanks for splitting it out. `EnginePipeline` reads well, consuming `self` in `stop()` to prevent use-after-stop is a nice touch, and the tightly-scoped macro is fine by me.

One thing worth recording, since the description says no behavior change: `stop()` does change error teardown from asymmetric to symmetric. The old arms joined only the sibling thread (e.g. `PlaybackError` joined capture, not playback); now every path joins both. I think that's an improvement, since the old code could silently orphan a thread that errored but didn't exit, and this surfaces it instead. Just noting it as an intentional tightening rather than a true no-op.

One small thing: `src/engine_pipeline.rs` is missing the GPL/MPL license header the other source files carry. Could you please add it? Happy to drop it in myself too, it's no big deal. Looks good to merge once that's in.

kvc0:
> One thing worth recording, since the description says no behavior change: stop() does change error teardown from asymmetric to symmetric. The old arms joined only the sibling thread (e.g. PlaybackError joined capture, not playback); now every path joins both. I think that's an improvement, since the old code could silently orphan a thread that errored but didn't exit, and this surfaces it instead. Just noting it as an intentional tightening rather than a true no-op.

Ah, thank you for pointing that out! I didn't mean to sneak that in there :facepalm: but I'm glad you approve at any rate!

> One small thing: src/engine_pipeline.rs is missing the GPL/MPL license header the other source files carry. Could you please add it?

I took care of it; thanks for the careful review.

HEnquist:
Great, thanks! Merging :)

PR #484, opened 2026-06-28
Engine tidying

kvc0:
separate signals handling, name thread:

I pulled the signals handling code out of engine; it has very few
dependencies, and I noticed that signal_hook is compiled in both
windows and unix builds. To help make it easier to statically
verify compilation on windows, I removed the conditional
compilation blocks and pulled the decision into a compile-time
decided const if {} block. This keeps the code there to satisfy
clippy w.r.t. symbol use.

This does have the side effect of also creating an active_config_path
pointer whether or not it will be used. It's benign and would be
dropped immediately on a windows non-websocket build. It also
names the signal thread. I kept the thread name below 16 characters
so it renders nicely on htop et al.

I added a couple more thread names while I was in here too.

### DIFF
--- a/src/engine.rs
+++ b/src/engine.rs
@@ -379,23 +379,26 @@ pub fn run_engine(engine_params: EngineConfig, logger: flexi_logger::LoggerHandl
         if let Some(fname) = &statefilename {
             let fname = fname.clone();
 
-            thread::spawn(move || {
-                loop {
-                    thread::sleep(Duration::from_millis(1000));
-                    match rx_state.recv() {
-                        Ok(()) => {
-                            debug!("saving state to {}", &fname);
-                            statefile::save_state(
-                                &fname,
-                                &active_config_path_clone,
-                                &processing_params_clone,
-                                &unsaved_state_changes,
-                            );
+            thread::Builder::new()
+                .name("statefile".to_string())
+                .spawn(move || {
+                    loop {
+                        thread::sleep(Duration::from_millis(1000));
+                        match rx_state.recv() {
+                            Ok(()) => {
+                                debug!("saving state to {}", &fname);
+                                statefile::save_state(
+                                    &fname,
+                                    &active_config_path_clone,
+                                    &processing_params_clone,
+                                    &unsaved_state_changes,
+                                );
+                            }
+                            Err(_) => break,
                         }
-                        Err(_) => break,
                     }
-                }
-            });
+                })
+                .expect("can spawn statefile thread");
         }
     }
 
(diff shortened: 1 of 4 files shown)


HEnquist:
CI caught the Windows break, and it's the conditional-compilation change. `cfg!(not(windows))` is a runtime bool, so both branches and all the file's `use` lines still compile on every platform. But `SIGHUP`, `SIGUSR1`, and `signal_hook::iterator` (`SignalsInfo`/`SignalOnly`) are all `#[cfg(not(windows))]` in signal-hook, so they don't exist on Windows. That's the `E0432` on the consts, and the `iterator` import failed the same way right after.

It compiles cleanly on Linux/macOS because the Windows-dead branch just gets optimized out there, which is why it slipped through locally.

The fix is to put the platform choice back at compile time while keeping the nice parts (named thread, separate file). The single spawn can stay, with the dispatch gated by `#[cfg]`:

```rust
thread::Builder::new().name("signal-monitor".into()).spawn(move || {
    #[cfg(not(windows))]
    monitor_unix_signals(active_path_thread, tx_command_thread, logger);
    #[cfg(windows)]
    monitor_windows_signals(tx_command_thread, logger);
})
```

and each `monitor_*` fn plus its imports gated by the matching `#[cfg]`, so the unix-only symbols never enter the Windows build. Heads up: this snippet is a suggestion I worked out with Claude and haven't actually compiled on Windows myself, so treat it as a direction rather than a verified patch. The exact shape that satisfies both the compiler and clippy may need a bit of fiddling.

Everything else here is good: the extraction, the thread naming on the statefile and processing/rayon threads, and the license header are all fine. Just this one needs to go back to `#[cfg]`.

One tiny cosmetic thing while you're in there: the `Arc<Mutex<parking_lot::RawMutex, Option<String>>>` param type reads more simply as `parking_lot::Mutex<Option<String>>`.

kvc0:
Yep, I missed that signal_hook internally uses not(windows), which broke the hopeful approach that made verifying windows seem more friendly. Anyway, I've addressed that break, the mutex alias, and I've updated the git history now that 483 is merged!

HEnquist:
The fix looks great, both the platform split and the mutex alias are clean.

One small thing, purely personal style and not a correctness issue: I'd rather keep imports at the top of the file and treat mid-function `use` as the exception, for cases where it genuinely can't live up top. These two can, with the gate riding along:

```rust
#[cfg(not(windows))]
use signal_hook::{
    consts::{SIGHUP, SIGUSR1, TERM_SIGNALS},
    iterator::{SignalsInfo, exfiltrator::SignalOnly},
};
```

Compiles identically, just moves the `signal_hook` imports out of `monitor_signals` and up to the top. Mind shifting them there? Then this is good to go.

kvc0:
I don’t mind a bit @HEnquist, thanks. It’s updated

PR #485, opened 2026-06-28
tighten rayon ownership

kvc0:
currently, rayon is initialized once as a global pool. If it fails,
the pipeline will not work. With this change, a warning will be
logged, and the application will fall back to as-is single-thread
filters processing.

Also, this change creates the pool per processing session. There's
only one, but it may be nice for library use to not have the global
rayon pool owned by camillalib.

By making rayon pool ownership clear and referential, it's easy to
see where and how each pool is used; and it makes it easier to
safely add more pools for other use cases in the future.

Multithread filter test with signal generator -> file through parallel
biquads appears to work fine, and tests still pass.

### DIFF
--- a/benches/pipeline.rs
+++ b/benches/pipeline.rs
@@ -219,7 +219,13 @@ fn build_pipeline(chunksize: usize, multithreaded: bool, with_conv: bool) -> Pip
     };
 
     let processing_params = Arc::new(ProcessingParameters::new(&[0.0_f32; 5], &[false; 5]));
-    Pipeline::from_config(conf, processing_params)
+    let filter_pool = camillalib::processing::build_processing_threadpool(
+        multithreaded,
+        conf.devices.worker_threads(),
+        conf.devices.chunksize,
+        conf.devices.samplerate,
+    );
+    Pipeline::from_config(conf, processing_params, filter_pool)
 }
 
 fn make_chunk(channels: usize, frames: usize) -> AudioChunk {
--- a/src/pipeline.rs
+++ b/src/pipeline.rs
@@ -23,6 +23,7 @@ use crate::filters::Filter;
 use crate::mixer;
 use crate::processors;
 use crate::processors::Processor;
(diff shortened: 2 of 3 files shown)


HEnquist:
Thanks for this. The good parts are the graceful fallback when the pool can't build, the named workers, and RT promotion via `start_handler` instead of `broadcast`.

But I don't want a per-pipeline pool, for two reasons.

Performance under load: the sweet spot for worker count is a few less than the core count, leaving room for capture, playback, processing, and the OS. I've benchmarked this and several users landed on the same result, more threads than cores is clearly bad. A single shared pool keeps total rayon parallelism within that budget no matter how many subsystems use it. Per-subsystem pools add up (filters now, maybe resampling later) and push past the core count, which only hurts when the machine is already busy, exactly when underruns happen.

Rebuild cost: the pool is rebuilt on every pipeline rebuild, inline on the processing thread between chunks. Spawning workers plus per-thread RT promotion (an rtkit/D-Bus call each on Linux), then joining the old workers, can be tens of ms on the thread that feeds playback, so a reload could glitch. It's also avoidable: `worker_threads`, `samplerate`, and `chunksize` all live in `devices`, and any change there forces a full restart, so within a session there's nothing to rebuild.

The nice parts still fit on a single pool: build one per processing session, own it explicitly, hand the same `Arc` to whoever needs it. That keeps the shared budget and your fallback/naming/start_handler wins, drops the rebuild, and avoids rayon's implicit global (nicer for the library case too).

kvc0:
that's a really good point about the processing session. Especially considering the library case. Thanks, I'm pulling it up to the processing loop.

(also rebasing the stacked pr so only 484 has to be ignored when reviewing commits. This is still draft and I'll update again once next4.2.0 is updated)

HEnquist:
Thanks, this is perfect. So this just needs the rebase once #484 lands, then I'll merge.

kvc0:
Cool, the pr stack should be resolved now!

HEnquist:
Thanks, merging!

PR #488, opened 2026-07-07
Enable realtime priority on linux

brainrake:
`audio_thread_priority` is a noop on linux without dbus (default) feature.
https://github.com/mozilla/audio_thread_priority/blob/c5c2d1cb5231215661ab68f3833beec4881d1604/src/lib.rs#L118

This change enables default features in audio_thread_priority, enabling dbus and rtkit on linux. It also adds dbus lib to CI build.

I would prefer a rlimit way to set rt prio, because it doesn't need dbus and can have higher limits. But dbus and rtkit is probably better suited for desktop systems.

I tried setting rt with systemd `CPUSchedulingPolicy` and `CPUSchedulingPriority` but it sets rt for all threads. Successfully prevents xruns under stress, a bit less than ideal.

I imagine there was a reason this feature was disabled (deps maybe?), so PR is draft.

### DIFF
--- a/.github/workflows/check_ws_docs.yml
+++ b/.github/workflows/check_ws_docs.yml
@@ -20,6 +20,9 @@ jobs:
       - name: Install ALSA development files
         run: sudo apt-get update && sudo apt-get install libasound2-dev -y
 
+      - name: Install DBus development files
+        run: sudo apt-get install libdbus-1-dev -y
+
       - name: Install Python dependencies
         run: pip install jinja2 pyyaml
 
--- a/.github/workflows/ci_test.yml
+++ b/.github/workflows/ci_test.yml
@@ -146,7 +146,7 @@ jobs:
       - name: Run cargo test for arm
         run: cross test --target arm-unknown-linux-gnueabihf
         env: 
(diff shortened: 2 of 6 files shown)


HEnquist:
Thanks for digging into this, and for the clear write-up. Your diagnosis is spot on: without the dbus feature the promotion is a no-op, so real-time priority has never actually worked on Linux.

The blocker for merging as-is is exactly the concern you raised. Enabling the dbus feature by default adds a hard runtime dependency on libdbus to every Linux binary, and rtkit needs its daemon running. CamillaDSP is very often run on minimal headless systems that have neither, so I don't want to require D-Bus for the default build.

I put up #489 with an approach that keeps both worlds:

- ALSA-only build promotes the thread directly with `pthread_setschedparam(SCHED_FIFO)`, no D-Bus.
- rtkit/D-Bus is kept for the `pulse-backend` and `pipewire-backend` builds, since those imply a desktop system where rtkit ships with the sound server anyway.

The headless case needs the process to be allowed to request RT priority (running as root, a `cap_sys_nice` capability, or an `RLIMIT_RTPRIO` limit via systemd `LimitRTPRIO` or limits.conf), which is documented in the ALSA backend docs there.

Would be great if you could take a look, and test it on your setup if you have a chance.

brainrake:
`pthread_setschedparam` FTW

PR #489, opened 2026-07-08
Enable real-time priority on Linux without D-Bus

HEnquist:
Real-time priority never actually worked on the Linux ALSA build: `audio_thread_priority` only promoted via rtkit over D-Bus, which is a no-op without the dbus feature, and that is how we build it.

- Bump `audio_thread_priority` to 0.36.0, which gained a native `SCHED_FIFO` path needing no D-Bus (mozilla/audio_thread_priority#45).
- New `rtkit` feature selects the D-Bus path, enabled automatically by `pipewire-backend`.
- `CAMILLADSP_RT_PRIORITY` env var sets the priority for the native path, default 10.
- Document the `RLIMIT_RTPRIO` setup for headless ALSA users in `backend_alsa.md`.
- macOS and Windows unchanged.

### DIFF
--- a/backend_alsa.md
+++ b/backend_alsa.md
@@ -282,6 +282,90 @@ For the gadget, the control can also indicate that the sample rate changed.
 When this happens, the capture can no longer continue and CamillaDSP will stop.
 The new sample rate can then be read by the `GetStopReason` websocket command.
 
+## Real-time priority
+CamillaDSP promotes its processing and audio threads to real-time priority (`SCHED_FIFO`), which helps
+avoid buffer underruns and dropouts under load. The default priority is 10, and it can be changed with
+the `CAMILLADSP_RT_PRIORITY` environment variable, see [Changing the priority](#changing-the-priority).
+
+The rest of this section applies to the plain ALSA-only build.
+There, priority is requested by calling `pthread_setschedparam` directly, without needing D-Bus or any
+running service, so it works on a minimal headless system.
+A build that includes the PipeWire backend instead uses `rtkit` over D-Bus to request the priority,
+which is the normal mechanism on a desktop system, and the setup below does not apply.
+An ALSA-only build can also be switched to `rtkit` by enabling the `rtkit` feature at build time.
+
+A process is only allowed to request real-time scheduling if it has permission to do so.
+Running as `root` works but is not recommended.
+The better option is to run CamillaDSP as a normal user and grant that user a real-time priority limit
+(`RLIMIT_RTPRIO`) that is at least as high as the priority CamillaDSP requests (10 by default).
+If the permission is missing, CamillaDSP still runs, but logs a warning that it could not get real-time
+priority.
+
+### Using systemd
+If CamillaDSP runs as a systemd service, set the limit directly in the unit file and run as a normal user:
+
+```ini
+[Service]
+User=camilladsp
+LimitRTPRIO=10
+```
+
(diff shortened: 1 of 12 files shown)


backend_alsa.md
```diff
@@ -282,6 +282,68 @@ For the gadget, the control can also indicate that the sample rate changed.
...
+If the permission is missing, CamillaDSP still runs, but logs a warning that it could not get real-time
+priority.
+
+### Using systemd
+If CamillaDSP runs as a systemd service, set the limit directly in the unit file and run as a normal user:
+
+```ini
+[Service]
+User=camilladsp
+LimitRTPRIO=95
```
brainrake:
```suggestion
LimitRTPRIO=10
```

backend_alsa.md
```diff
@@ -282,6 +282,68 @@ For the gadget, the control can also indicate that the sample rate changed.
...
+
+This is the most self-contained option, as the limit is part of the unit file.
+
+### Using PAM limits
+For a login session, or when not using systemd, grant the limit through PAM.
+Create a drop-in file:
+
+```
+# /etc/security/limits.d/95-camilladsp.conf
+@audio   -   rtprio   95
```
brainrake:
```suggestion
@audio   -   rtprio   10
```

src/utils/rt_priority.rs
```diff
@@ -0,0 +1,132 @@
...
+#[cfg(all(target_os = "linux", not(feature = "rtkit")))]
+pub use native::{demote_current_thread_from_real_time, promote_current_thread_to_real_time};
+
+#[cfg(all(target_os = "linux", not(feature = "rtkit")))]
+mod native {
+    use std::fmt;
+
+    /// Real-time priority to request for the SCHED_FIFO policy. Matches the default used by rtkit
+    /// (`RT_PRIO_DEFAULT`), which is well below the priority of critical kernel threads.
+    const RT_PRIORITY: libc::c_int = 10;
```
brainrake: I wish there was a way to configure this. Or maybe get the limit and use that as the value.

brainrake: Anecdotes about how rtkit defaults are too low:
https://github.com/rerdavies/pipedal/discussions/99

brainrake (changes requested):

brainrake:
Great stuff! Should have been upstream :upside_down_face:
Love the feature flag, allows any combination of rtkit/rlimit and backend.
I need a bit of time to test both rt paths.

HEnquist:
Thanks for the review, and the pipedal link.

Made the priority configurable with a `CAMILLADSP_RT_PRIORITY` env var (1-99, default 10) instead of a config option, so it stays out of the schema and drops into a systemd unit easily. Also took your doc edits, the limit examples now use 10.

I added a doc caveat about keeping the priority below the audio IRQ thread (~50), above that you preempt the threads feeding the device and cause dropouts. That's also why I used an env var rather than deriving from `RLIMIT_RTPRIO`. Keeping the default at 10.

HEnquist:
I made a PR here: https://github.com/mozilla/audio_thread_priority/pull/45
Let's wait a little, if that gets merged and released soon, the code here can be made much simpler.

brainrake:
Wow, looks like your PR was merged and audio_thread_priority 0.36.0 released! :tada: 
I need to do some rt testing so I will revisit this branch.

brainrake:
Nice!

PR #494, opened 2026-07-13
Add native PureRoad Airwindows character processor

lighthx:


HEnquist:
I am not going to merge this. The main reason is scope, not implementation. CamillaDSP is a signal correction and dynamics tool. It handles things like crossovers, room correction, and compression, limiting, and soft clipping. It is not meant to be a place for tone shaping or coloration effects. I have turned down tube emulation for the same reason before. A character or tape emulation processor does not fit that scope, no matter how it is implemented.

Even if this fit the scope, the current implementation would still be a problem for me. Adding a C++ toolchain dependency changes the build story for every target CamillaDSP supports, including the ARM cross compiles many users rely on. Vendoring a snapshot of someone else's DSP code into this repository also creates a long term license and maintenance burden I am not willing to take on, even under a permissive license like MIT.

A pure Rust reimplementation of the same algorithms would not change my answer. The issue is not the C++, it is that this kind of processor is not something I want CamillaDSP to include at all.

I am closing this PR.

PR #504, opened 2026-07-25
Add LookaheadLimiter processor

HEnquist:
Multichannel version of the LookaheadLimiter filter, with monitor and process channels like the Compressor.

- Peak detection on `monitor_channels`, using the largest amplitude of the monitored channels
- Gain reduction applied to `process_channels`
- All channels are delayed by the attack time so they stay aligned, `delay_processed_only` skips the delay on the unprocessed ones
- Limiter algorithm extracted into a shared `LookaheadGain` struct used by both the filter and the processor
- README and CHANGELOG updated

@brainrake you wrote the lookahead limiter filter, so it would be great to get your eyes on this.


HEnquist:
@brainrake merging now so it doesn't rot against next5, shout if you spot anything and I'll fix it in a follow-up.

brainrake:
This looks good! Thanks a lot! I'll try it out.

Issue #350, opened 2024-07-01
Suggestion: Publish camilladsp to crates.io


ameknite:
In the rust community is common to upload binaries to [crates.io](https://crates.io/) so it can be easier to install using:

```cargo install camilladsp```

or download the binaries from github using [`cargo binstall`](https://github.com/cargo-bins/cargo-binstall)

```cargo binstall camilladsp```

This will also make easier to update the program with [cargo-update](https://github.com/nabijaczleweli/cargo-update):

```cargo install-update camilladsp```

if you like the suggestion, I could help with the github workflow

HEnquist:
I'm publishing a few libraries on crates.io, for example https://crates.io/crates/rubato

I have considered publishing also camilladsp on crares.io but what has kept me from doing it is that then all dependencies must be published there too. That is normally the case, but a few times I have used forks of some libraries to get a new camilladsp version out without having to wait for the updated library. Such a version could not be published to crates.io so the version there would lag behind.

ameknite:
Oh, I didn't know about that. But if you want to keep the same versions in sync, I think you can actually publish the forks with a different name to crates.io, or even vendor the crates. But, I understand it can be annoying to keep updating the forks.

The easiest solution would be to just publish to crates.io and add a disclaimer in the readme that the version could be left behind. I think it is better to have that version than none at all, but it would be up to you.

ameknite:
It looks like right now camilladsp is not using any external crates. So at least the current version can be published.

HEnquist:
Another point is that I think crates.io is primarily used by rust developers, mostly to find libraries and occasionally some tools they need for some project. Would anyone go looking there for something like camilladsp?

I haven't tried but `cargo install` should also work without publishing to crates.io, with the `--git` option: https://doc.rust-lang.org/cargo/commands/cargo-install.html#synopsis

ameknite:
Yeah, I just thought it was a nice to have, maybe something like homebrew might be better.

HEnquist:
Yes homebrew is great. It would be neat if someone made and submitted a formula 😉

PR #515, opened 2026-08-30
Publish to crates.io, and support cargo install / cargo binstall

illutr0n:
Picking up #350. This adds what is needed to publish CamillaDSP on crates.io, so it can be installed and updated with:

```sh
cargo binstall camilladsp   # downloads the release binary
cargo install camilladsp    # compiles from source
cargo install-update camilladsp  # via cargo-update
```

Nothing is published by this PR on its own — it only makes the crate publishable. Feel free to close it if you'd rather not be on crates.io; the name `camilladsp` is still unregistered today, so it is at least worth claiming.

### What's in it

**`Cargo.toml`**
- Renamed the package from `CamillaDSP` to lowercase `camilladsp`. crates.io names are case sensitive for `cargo install`, so this is what makes `cargo install camilladsp` work, and it matches the executable name. The `[lib] name = "camillalib"` and `[[bin]] name = "camilladsp"` targets are set explicitly, so nothing about the build output changes.
- Added `repository`, `homepage`, `readme`, `keywords` and `categories`, which is what crates.io wants for a decent listing page.
- Added an `exclude` list, so the published archive contains the sources, benches, testdata, example configs and docs, but not the images, CI files, Dockerfiles, `web/` or `testscripts/`. `testdata/` is kept, so `cargo test` works on the published crate.
- Added `[package.metadata.binstall]` mapping the seven target triples to the release asset names from `publish.yml`. The `.tar.gz` and `.zip` archives contain the executable at the top level, so `bin-dir` is just the binary name. This means `cargo binstall` downloads the same binary as the Releases page rather than compiling.

**`src/bin.rs`**
- `crate_name!()` would now expand to the lowercase `camilladsp` in the `--help` text. Replaced it with a new `APP_NAME` const, which also replaces the `"CamillaDSP"` literal that was already passed to `Command::new`. The `--help` output is unchanged.

**`.github/workflows/publish_crates_io.yml`** (new)
- Runs `cargo publish` when a release is created, and can also be triggered manually. It needs a `CARGO_REGISTRY_TOKEN` repository secret, from https://crates.io/settings/tokens — ideally a token scoped to `publish-update` for this crate only. The first publish has to be done by hand, since the token can only be scoped to a crate that already exists.

**`.github/workflows/ci_test.yml`**
- Added a job that runs `cargo package`, so a broken package is caught in CI rather than in the middle of a release.

### Notes

- Regarding your point in #350 about forks: all dependencies currently resolve to crates.io releases, the fork lines are commented out, so 4.1.3 is publishable as is. If a fork is ever needed again, that release just gets skipped on crates.io, and the workflow can be left to fail or the job skipped for that release.
- `git_version!` already has `fallback = "unknown"`, so the version string degrades gracefully when built from the crates.io tarball with no git repo present.
- I have not added a CHANGELOG entry, since there is no unreleased section — happy to add one wherever you prefer.
- I don't have a Rust toolchain on this machine, so I have not built this locally; CI on this PR is the first real check. Let me know if anything needs fixing.

### DIFF
--- a/.github/workflows/ci_test.yml
+++ b/.github/workflows/ci_test.yml
@@ -212,3 +212,22 @@ jobs:
 
 
 
+
+  package:
+    name: Check crates.io package
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout sources
+        uses: actions/checkout@v5
+
+      - name: Install stable toolchain
+        uses: dtolnay/rust-toolchain@stable
+
+      - name: Update package list
+        run: sudo apt-get update
+
+      - name: Install Alsa devel
+        run: sudo apt-get install libasound2-dev -y
+
+      - name: Build the crate package
+        run: cargo package
--- a/.github/workflows/publish_crates_io.yml
+++ b/.github/workflows/publish_crates_io.yml
@@ -1,0 +1,27 @@
+name: Publish to crates.io
+
+on:
+  release:
+    types:
+      - created
+  workflow_dispatch:
+
+jobs:
+  crates_io:
+    name: Publish to crates.io
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout sources
+        uses: actions/checkout@v5
+
+      - name: Install stable toolchain
+        uses: dtolnay/rust-toolchain@stable
+
+      - name: Update package list
+        run: sudo apt-get update
+
+      - name: Install Alsa devel
+        run: sudo apt-get install libasound2-dev -y
+
+      - name: Publish
+        run: cargo publish --token ${{ secrets.CARGO_REGISTRY_TOKEN }}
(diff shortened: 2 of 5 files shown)


illutr0n:
Closing this — on reflection I'd rather pursue the Homebrew formula you mentioned in #350 than the crates.io route. Sorry for the noise.

HEnquist:
Thanks for putting this together, it wasn't noise at all.

camilladsp is going up on crates.io, probably around the 5.0 release. The library part is what I'm focussing on there though, rather than `cargo install`, so the binstall side isn't really the driver for me. The 5.0 branch already has a fair bit of the packaging metadata in place.

Good catch on `crate_name!()` leaking into the `--help` header, I'd have walked straight into that one.

A Homebrew formula would still be very welcome if you want to go that way, either homebrew-core or a tap of your own, whichever suits you.

PR #533, opened 2026-09-21
Add linear-phase FIR crossover filter type

Vasiliy-Bondarenko:
## What

A new filter type, `Crossover`. It is a linear-phase FIR crossover that is defined by a frequency and a slope, instead of by a file with coefficients:

```yaml
filters:
  low:
    type: Crossover
    parameters: { type: Lowpass, freq: 2000, slope: 48 }
  high:
    type: Crossover
    parameters: { type: Highpass, freq: 2000, slope: 48 }
```

The slope is given in dB per octave, as a multiple of 6 from 12 to 96. The coefficients are calculated when the config is loaded, and the filtering itself reuses the existing `FftConv`.

## Why

Using a linear-phase crossover today means designing the coefficients in an external tool, exporting a file per band and per samplerate, and keeping those files in sync with the config. This makes a crossover a matter of two numbers, and the coefficients follow the samplerate automatically.

## How it works

The lowpass magnitude is `1/(1+r(x))` and the highpass is `r(x)/(1+r(x))`, where `x` is the prewarped normalized frequency and `r(x) ~ x^n` with `n = slope/6`. Both bands are -6 dB at the crossover frequency, and because the highpass is the delta minus the lowpass, the two bands sum to a pure delay with a flat magnitude response.

The FIR length is not fixed. It is searched so that the truncation error, bounded by the tail sum of the impulse response, stays below -100 dB, and a lowpass and highpass with the same settings always get the same length.

## Latency alignment

Since the length depends on the settings, different channels can end up with different latencies. A three-way setup has two crossovers in series on the mid channel but only one on the woofer and the tweeter. A mismatch of even a few samples breaks the summation, so `Pipeline::from_config` now tracks the crossover latency per channel and delays the channels that are behind. This is done before every mixer and processor step, and at the end of the pipeline. Manually added `Delay` filters are not touched. The total latency is written to the log.

Because the latency can change with the parameters, `config_diff` returns `ConfigChange::Pipeline` for any change to a `Crossover` filter, so the pipeline is rebuilt and the alignment recalculated.

## Testing

Five new unit tests: the bands sum to a pure delay, -6 dB at the crossover frequency, the response matches the target formula, lowpass and highpass have equal latency, and config validation. Plus `known_latencies`, a fixed table of expected latencies that also exists in the GUI-side implementation (see below), so the two cannot drift apart unnoticed.

Verified end to end by running an impulse through the binary in a three-way configuration (80 Hz/48 dB per octave and 2500 Hz/96 dB per octave), plus one channel with no filters: all four channels peak on the same sample, and the three bands sum back to the original impulse with a residual of -158 dB.

`cargo test`, `cargo clippy --all-targets` and `cargo fmt` are clean.

## Related

The GUI needs to show the latency without a running engine, so the same algorithm exists in pycamilladsp-plot. That copy, the backend endpoint and the GUI form are separate pull requests, and each repo carries a note about the duplication:

- HEnquist/pycamilladsp-plot#41: validation, evaluation and latency
- HEnquist/camillagui-backend#95: `/api/crossoverlatency`
- HEnquist/camillagui#173: filter form and latency display

I am happy to adjust the naming, the -100 dB tolerance or the alignment rules if you would prefer them different.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


HEnquist:
Thanks for this, and that's clearly a lot of work. It looks well put together and I have no reason to think it doesn't do what you describe.

Unfortunately I'm going to have to say no. For anything non-trivial, and especially new functionality, please open an issue first next time. This one would have been answered in a day, and it would have saved you four pull requests.

FIR filter design is an area I have deliberately stayed out of. It has come up a few times over the years in the diyAudio thread and I have said no each time. It's a huge field, and there are dedicated tools for it. The README points at several already, under "Measurement and filter generation tools". Anything built into CamillaDSP would be limited and opinionated next to those, and there's no principled place to stop once the first one is in. (The Biquad filters are different, those come from a closed set of textbook formulas with nothing to tune.)

CamillaDSP already has Linkwitz-Riley and Butterworth crossovers of any order, as IIR. This adds a single fixed shape that is neither, so the obvious next request is a linear-phase version of the alignments that are already there, and then Bessel, and then brickwall, and then minimum phase, and mixed phase somewhere between linear and minimum. Then the design parameters themselves: there is no window function to choose, no way to cap the length when you have a latency budget, and the -100 dB tolerance that decides the length is hardcoded. All reasonable things to want control over, and the filter config starts turning into a filter design UI. And roughly ten minutes after release someone will sensibly ask for room correction from .frd files. Every one of those requests is reasonable on exactly the same grounds as this PR, and I'd have no good answer to any of them. Chasing that as a side feature of a DSP engine means doing it badly and never catching up. The more immediate limit is simply time. Everything here has to stay within what one person can maintain alone, and the current scope is already at that limit.

There's also a structural limit. The latency here is just the filter length, which only works because the impulse response is symmetric by construction, and the whole channel alignment in the pipeline rests on that. Minimum phase has no single latency to report, only frequency-dependent group delay, and mixed phase, which is what people usually actually want since it trades pre-ringing against delay, has none either. Linear phase is the one point in that design space where the question has an easy answer, and this design can only ever represent that point. So it wouldn't extend, it would have to be redone, and the alignment wouldn't fail loudly, it would just quietly align channels that aren't aligned.

One more thing, and this one is just my opinion. Linear phase applied only to the crossover doesn't buy much on its own. The drivers have their own rolloffs and phase shifts, and those dominate what actually arrives at the listening position. The tests here show that lowpass and highpass sum to a pure delay, and they do, electrically. Acoustically that identity is gone as soon as real drivers are involved. So you pay the full cost of linear phase, a lot of latency and pre-ringing, for a guarantee that doesn't survive the speaker. The tools that make linear phase worth having get there by measuring the drivers and correcting the whole chain at once, with the crossover as one part of it.

rePhase is the obvious tool for this. It builds fully linear-phase crossovers with arbitrary slopes, free, and lets you correct the drivers in the same filter. The README has pointed at it for a long time, in the section on generating FIR coefficients. Export a .dbl at the right rate and Conv reads it directly.

If you want something more automated than that, a standalone generator producing coefficient files is the natural form, and most of what you've written carries over. Generate one set per samplerate and name the files with the rate, and the `$samplerate$` token in the Conv filename picks the right one automatically. A fork is of course also fine.

